PR #117, opened 2015-02-03
allow codec to be none for old versions of ffmpeg

aestrivex:
The save_movie() function calls ffmpeg with the -c argument which breaks very old versions of ffmpeg such as versions which can be found on old installations of RHEL/CentOS. This PR allows the omission of the -c argument when the codec kwarg is None.

### DIFF
--- a/surfer/utils.py
+++ b/surfer/utils.py
@@ -703,10 +703,10 @@ def ffmpeg(dst, frame_path, framerate=24, codec='mpeg4'):
     frame_dir, frame_fmt = os.path.split(frame_path)
 
     # make the movie
-    if codec is None:
-        cmd = ['ffmpeg', '-i', frame_fmt, '-r', str(framerate), dst]
-    else:
-        cmd = ['ffmpeg', '-i', frame_fmt, '-r', str(framerate), '-c', codec, dst]
+    cmd = ['ffmpeg', '-i', frame_fmt, '-r', str(framerate)]
+    if codec is None: 
+        cmd += ['-c', codec]
+    cmd += [dst]
     logger.info("Running FFmpeg with command: %s", ' '.join(cmd))
     sp = subprocess.Popen(cmd, cwd=frame_dir, stdout=subprocess.PIPE,
                           stderr=subprocess.PIPE)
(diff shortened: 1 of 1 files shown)


surfer/utils.py
```diff
@@ -701,7 +703,10 @@ def ffmpeg(dst, frame_path, framerate=24, codec='mpeg4'):
     frame_dir, frame_fmt = os.path.split(frame_path)
 
     # make the movie
-    cmd = ['ffmpeg', '-i', frame_fmt, '-r', str(framerate), '-c', codec, dst]
+    if codec is None:
+        cmd = ['ffmpeg', '-i', frame_fmt, '-r', str(framerate), dst]
+    else:
+        cmd = ['ffmpeg', '-i', frame_fmt, '-r', str(framerate), '-c', codec, dst]
```
Eric89GXL: Would be more DRY to do something like:
```
cmd = ['ffmpeg', '-i', frame_fmt, '-r', str(framerate)]
if codec is not None:
    cmd += ['-c', codec]
cmd += [dst]
```

aestrivex: OK, done that way.

Eric89GXL:
I assume you've tested this and it works for your code?

aestrivex:
Yes it works, it is very slow to update though for brains with a large number of vertices.

I tried it on fsaverage5 and it was quite a lot faster.

Eric89GXL:
I wonder where the bottleneck is... feel free to open an improvement PR if you find a better way. We spent a bit of time getting the API right, but changing the under-the-hood mechanics shouldn't be too bad from an API standpoint. Maybe MoviePy would speed things up.

aestrivex:
Part of the bottleneck may be in repeatedly saving images, but I think a far larger bottleneck is in the large amount of interpolation done for a large amount of vertices (which the user can adjust currently).

I have some other code (which is not trivially portable) which captures movies from mayavi by using ffmpeg to do screen capture rather than repeatedly taking screenshots. I tried both methods, and found that doing the direct screen capture does save a lot of overhead. My implementation might have been inefficient though, I will try it here and see if it does any better.

Eric89GXL:
We do want to keep compatibility with other platforms, though, so I'm not sure that would end up working.

Issue #120, opened 2015-02-27
Get rid of the config file?


mwaskom:
I almost always find the `config_opts` dictionary to be a huge pain. The lack of tab completion in IPython means I always have to hunt to find the right name for the keys, and it's weird to pass an extra dictionary of keyword arguments.

I can't even remember why those options were implemented this way, though I suspect it made reading them from the config file easier. I would like to get rid of the config_opts dictionary and make its entries first-class keyword arguments in `Brain`. (We can keep `config_opts` for a release cycle or two and fire a warning when people use it, to avoid too much disruption).

But more generally, it's worth thinking about just killing the config file itself. I've come to really dislike this approach to customization because it means the scripts/notebooks you write aren't fully reproducible without metadata, and there's no indication in the code that the metadata existed at the time it was executed. (i.e., the script won't even crash, you'll just get a result that looks different).

I also suspect that the config file is relatively underutilized.

Would anyone object to just getting rid of the config file and making everything be specified in the `Brain` constructor?

mwaskom:
Soliciting comment from @Eric89GXL, @agramfort, and anyone else who wants to weigh in on whether they feel the config file is an integral part of their pysurfer workflow and is worth keeping.

agramfort:
+1

and I personally don't use the config file

Eric89GXL:
+1, same here (including the annoyances with working with `config_opts`)

mwaskom:
Closed with #121

PR #121, opened 2015-03-01
Remove config_opts and config file

mwaskom:
This addresses #120

Values that were previously selected using a ``config_opts`` dictionary in the ``Brain`` constructor are now keyword arguments. This should make it easier to use tab-completion in IPython, and will generally simplify your scripts. To ease transition, a ``config_opts`` argument will still be caught, and its entries will be used (overriding the keyword arguments), but it will be removed in a future release.

The ability to set default values in a config file has been removed. While convenient, this approach encourages code that is not fully reproducible. While existing code that was written expecting a config file will still run, the visualization will fall back to default values. These should be updated directly in your plotting scripts.

While not directly related to these changes, this  PR also adds a new feature: the ``cortex`` keyword argument can now be a mayavi colormap name or a ``(colormap, min, max, reverse)`` tuple for full control.


mwaskom:
I think I've got everything and the tests pass on my machine, but please take a look.

doc/documentation/custom_viz.rst
```diff
@@ -12,7 +12,17 @@ have to delve into the underlying engine to get a different look.
...
+The display background can take any valid matplotlib color (i.e.,
+it can be a tuple of rgb values, an rgb hex string, or a named HTML
+color.
+
+Changing the display size
+-------------------------
+
+The default display window is 800px by 800px, but this is also
+customizable. You can set the size of a square window with the ``size``
+parameter or make a rectangular window with the ``width`` and ``height``
```
Eric89GXL: While we're at it, how about we deprecate `size`, too? Just about every other viz package uses `width` and `height`, and it's more pythonic to have one way of doing it.

Eric89GXL:
Other than my one idea, LGTM +1 for merge

doc/documentation/custom_viz.rst
mwaskom: I agree that in general it feels like a clunky API, and that just taking width and height is more common (although usually that's `size = (width, height)`, I feel like).

Eric89GXL: Ahh, right. So maybe deprecate `width` and `height` and allow `size` to be an int or tuple/list of two int?

mwaskom: I think this is probably the best API, although I think it produces the ugliest code for handling backwards compatibility properly.

agramfort: +1 for size with tuple as figsize in matplotlib

agramfort:
besides diff looks good

mwaskom:
OK I think that will do the trick.

mwaskom:
damn @coveralls u strict

Eric89GXL:
We can opt to ignore it, though :) Good to go from your end, then?

Eric89GXL:
Oh right "yes" because of previous comment, merging. Thanks for cleaning this up.

agramfort:
Yes

PR #122, opened 2015-03-02
Fix errors that were crashing doc build

mwaskom:
GH-#115 removed surfer.io.read_label but did not change the example
that was referencing it. I'm surprised that got through Travis--
does Travis not build the docs?

While fixing this I also noticed a couple of things that had gotten
out of date and fixed an oversight from the config file removal.


Eric89GXL:
Travis does not build the docs, no. Feel free to open a PR for that :) It could be another worker condition.

Eric89GXL:
LGTM, +1 for merge once the CIs are happy

PR #125, opened 2015-03-26
fix severe bug with codec being None

aestrivex:
Sorry there was a slight bug with a style correction that was made here.


agramfort:
nice catch thx !

PR #126, opened 2015-03-26
Fix sample floatingpoint offby1

aestrivex:
I noticed that `save_movie()` was creating movies slightly shorter than the desired length -- a 6 second movie would become a 5.8 seconds movie, for instance.

I think this is primarily an issue in ffmpeg which would be almost impossible to solve, but I also noticed a minor floating point instability in the calculation of the number of samples in my data.


agramfort:
nice catch ** 2 !

agramfort:
thanks heaps !

PR #129, opened 2015-05-05
ENH Brain:  allow time_label to be function

christianbrodbeck:
This should solve the problem of displaying time in ms while representing it in seconds (which is assumed by `Brain.save_movie()` when determining time dilation). To display in ms use `time_label=lambda x: '%s ms' % int(round(x))`.


agramfort:
LGTM

Eric89GXL:
Thanks @christianbrodbeck!

PR #130, opened 2015-05-05
add bitrate option to ffmpeg call

aestrivex:
I noticed the movies I was creating from pysurfer were sometimes very low quality.

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -2122,7 +2122,8 @@ class Brain(object):
         return out
 
     def save_movie(self, fname, time_dilation=4., tmin=None, tmax=None,
-                   framerate=24, interpolation='quadratic', codec='mpeg4'):
+                   framerate=24, interpolation='quadratic', codec='mpeg4',
+                   bitrate='750k'):
         """Save a movie (for data with a time axis)
 
         .. Warning::
@@ -2150,6 +2151,9 @@ class Brain(object):
             'cubic', default 'quadratic').
         codec : str
             Codec to use with ffmpeg (default 'mpeg4').
+        bitrate : str | float
+            Bitrate to use to encode movie. Can be specified as number (e.g.
+            64000) or string (e.g. '64k'). Default value is 750k
 
         Notes
         -----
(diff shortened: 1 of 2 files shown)


christianbrodbeck:
Implementation looks good, I wonder if it's a good idea to set the default as constant, i.e. to have the same bitrate whether you are making a high resolution movie with 6 brains or a small one with a single brain. Maybe the default should still be `None` and that should call ffmpeg without an argument?

surfer/viz.py
```diff
@@ -2192,7 +2196,7 @@ def save_movie(self, fname, time_dilation=4., tmin=None, tmax=None,
         fname_pattern = os.path.join(tempdir, frame_pattern)
         self.save_image_sequence(time_idx, fname_pattern, False, -1, -1,
                                  'current', interpolation=interpolation)
-        ffmpeg(fname, fname_pattern, framerate, codec)
+        ffmpeg(fname, fname_pattern, framerate, codec=codec, bitrate=bitrate)
```
christianbrodbeck: could just be `ffmpeg(fname, fname_pattern, framerate, codec, bitrate)` but that's a detail

aestrivex:
I picked the default value to be approximately the same value (i.e., movies
that gave me similar performance) to setting no argument. It turns out that
according to the ffmpeg documentation the default bitrate is 200k and not
750k as I specified, which is still pretty poor quality. So I would prefer
to change the default value to that but keep it in the code to be explicit.

On Tue, May 5, 2015 at 9:32 PM, Christian Brodbeck <notifications@github.com

christianbrodbeck:
So you're saying 750k looks approximately the same as 200k?

aestrivex:
For the movies and resolution I made, they are pretty comparable. I cut it
up to 2m and it's considerably better. But anyway it seems reasonable to
use ffmpeg's default and let the user adjust it.

christianbrodbeck:
I think the default could go either way, sometimes you want movies that are higher quality and sometimes you want to save several movies for reference without clogging your heart drive... the increase in bitrate would have to be justified by improved quality though, so if 750k looks the same as 200k maybe it's not an ideal default...

christianbrodbeck:
What would be useful would be a note in the documentation on what bitrate to use to get good quality

Eric89GXL:
I have a feeling users will want good movies more often than they'd want crappy ones, and at least by default it would be nice to produce something that looks good. So +1 for using 2M or whatever looks good by default for me.

christianbrodbeck:
Yeah I agree with a default that looks good

agramfort:
+1 for merge too.

aestrivex:
Okay, I did a little bit of testing and increased the default value to 1M which seems to be a consistently good value (ffmpeg's behavior is a little strange). My movies are pretty high resolution though so I think for many people a lower bitrate would be acceptable, but the filesize of the movies I generate with this is not so unreasonable (on the order of megabytes or tens of megabytes if the movie is long) so I don't think setting the default here is a big deal.

Eric89GXL:
Sounds good, I'll merge once Travis comes back happy

Eric89GXL:
Thanks @aestrivex

PR #131, opened 2015-05-06
ENH:  Time-label consistency over time

christianbrodbeck:
Mayavi's default bahavior is to change the text size when the number of characters changes, which looks  strange in a movie (see [before](https://www.dropbox.com/s/j551izzp9o405d9/example_before.mov?dl=0)). This PR tries to fix this tentatively by using a constant font size and right-aligning the label (see [after](https://www.dropbox.com/s/9xjjwc7gj10yswp/example_after.mov?dl=0)). I don't know how much of this should be exposed in `Brain.add_data()`, maybe at least the font size?

### DIFF
--- a/examples/save_movie.py
+++ b/examples/save_movie.py
@@ -27,8 +27,8 @@ for hemi in ['lh', 'rh']:
     data = stc['data']
     times = np.arange(data.shape[1]) * stc['tstep'] + stc['tmin']
     brain.add_data(data, colormap='hot', vertices=stc['vertices'],
-                   smoothing_steps=10, time=times, time_label='%0.3f s',
-                   hemi=hemi)
+                   smoothing_steps=10, time=times, hemi=hemi,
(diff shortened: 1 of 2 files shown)


Eric89GXL:
This will break backward compatibility a bit, since it will change how the labels get plotted. I think I'm okay with it in this case, though. @agramfort WDYT?

christianbrodbeck:
Yes I was considering exposing the old default as an argument, but since the old default had this strange effect whenever the number of digits changed maybe something more useful can be exposed...

agramfort:
it's more a bug fix so I am fine with it. You use more these features than me anyway :)

christianbrodbeck:
Ok, I exposed the font size for the label. Do you want more options?

Eric89GXL:
Looks good, we can add more later if we need them

Eric89GXL:
Thanks again @christianbrodbeck

Issue #102, opened 2014-04-08
ENH: avoid having to run FreeSurferEnv.sh


GaelVaroquaux:
In the different places where pysurfer calls freesurfer commands, the following idiom would enable working without sourcing "FreeSurferEnv.sh" before opening Python:

<pre>
subprocess.call(
      'source $FREESURFER_HOME/FreeSurferEnv.sh; mri_vol2surf --src %s --out %s.left.mgz --srcreg $FREESURFER_HOME/average/mni152.register.dat --hemi lh'
      % (tmp_file.name, outfile), shell=True, executable="/bin/bash")
</pre>

It would be trivial to change the various calls to "subprocess.call" like this, and thus make things more user-friendly.

effigies:
Just noticed this is still open. @GaelVaroquaux did #132 cover your use case?

Eric89GXL:
Closing due to #132, feel free to reopen if this still occurs

PR #132, opened 2015-05-07
Check environment and run FreeSurferEnv.sh if necessary

effigies:
Addresses #102.

Check `os.environ` and run `$FREESURFER_HOME/FreeSurferEnv.sh` if not the most recent script to edit the `PATH`. Parse the resulting environment and pass to `mri_vol2surf`, avoiding modifying the base process environment.

### DIFF
--- a/surfer/io.py
+++ b/surfer/io.py
@@ -1,7 +1,7 @@
 import os
 from tempfile import mktemp
 
-from subprocess import Popen, PIPE
+from subprocess import Popen, PIPE, check_output
 import gzip
 import numpy as np
 import nibabel as nib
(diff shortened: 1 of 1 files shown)


Eric89GXL:
LGTM. I assume you've tried it and it works?

effigies:
Yes, run through all new paths.

effigies:
Though, it would mistake a manual `PATH="$FREESURFER/bin:$PATH"` for having previously run `FreeSurferEnv.sh`. If that's a problem, I could use a different check.

Eric89GXL:
I think that's an okay corner case to address later if need be

agramfort:
if this works I am happy

mwaskom:
This won't work on windows, but I am guessing that is not an issue because this whole function won't work on windows anyway?

effigies:
That was my assumption.

Eric89GXL:
Thanks @effigies!

PR #133, opened 2015-05-10
FIX Brain.save_movie():  prevent greater than possible frame times

christianbrodbeck:
np.arange() sometimes returns values larger than `stop` due to floating 
point round-off, which becomes relevant with large `time_dilation`
values.

I hope that's the last fix for determining movie times :)

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -1,5 +1,6 @@
 import os
 from os.path import join as pjoin
+import sys
 from tempfile import mkdtemp
 from warnings import warn
 
@@ -2178,19 +2179,16 @@ class Brain(object):
             raise ValueError("tmin=%r is smaller than the first time point "
                              "(%r)" % (tmin, self._times[0]))
 
-        if tmax is None:
-            tmax = self._times[-1]
-        elif tmax >= self._times[-1]:
-            raise ValueError("tmax=%r is greater than the latest time point "
-                             "(%r)" % (tmax, self._times[-1]))
-
         # find indexes at which to create frames
         tstep = 1. / (framerate * time_dilation)
-        if np.allclose((tmax - tmin) % tstep, 0):
-            tstop = tmax + tstep / 2.
-        else:
-            tstop = tmax
-        times = np.arange(tmin, tstop, tstep)
+        if tmax is None:
+            tmax = self._times[-1]
+        elif tmax > self._times[-1]:
+            raise ValueError("tmax=%r is greater than the latest time point "
+                             "(%r)" % (tmax, self._times[-1]))
+        times = np.arange(tmin, tmax + sys.float_info.epsilon, tstep)
+        while times[-1] > tmax:
+            times = times[:-1]
         interp_func = interp1d(self._times, np.arange(self.n_times))
         time_idx = interp_func(times)
 
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -2187,7 +2187,7 @@ class Brain(object):
             raise ValueError("tmax=%r is greater than the latest time point "
                              "(%r)" % (tmax, self._times[-1]))
         times = np.arange(tmin, tmax + sys.float_info.epsilon, tstep)
-        while times[-1] > tmax:
+        if times[-1] > tmax:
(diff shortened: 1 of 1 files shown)


surfer/viz.py
```diff
@@ -2178,19 +2179,16 @@ def save_movie(self, fname, time_dilation=4., tmin=None, tmax=None,
...
-        # find indexes at which to create frames
-        tstep = 1. / (framerate * time_dilation)
-        if np.allclose((tmax - tmin) % tstep, 0):
-            tstop = tmax + tstep / 2.
-        else:
-            tstop = tmax
-        times = np.arange(tmin, tstop, tstep)
+        times = np.arange(tmin, tmax + sys.float_info.epsilon, tstep)
+        while times[-1] > tmax:
+            times = times[:-1]
```
agramfort: don't like while here.

times = times[times < tmax]

?

christianbrodbeck: I could use `if`, I don't assume there are ever more than 1 values larger than tmax. `times = times[times < tmax]` seems less efficient because you create and copy arrays in every iteration

agramfort: this is likely to be negligible and I have seen too many while loops
freezing computers... I tend to avoid them.

christianbrodbeck: Well this one will stop at the latest when `times` reaches length 0 :)

surfer/viz.py
```diff
@@ -2178,19 +2179,16 @@ def save_movie(self, fname, time_dilation=4., tmin=None, tmax=None,
...
-
-        # find indexes at which to create frames
-        tstep = 1. / (framerate * time_dilation)
-        if np.allclose((tmax - tmin) % tstep, 0):
-            tstop = tmax + tstep / 2.
-        else:
-            tstop = tmax
-        times = np.arange(tmin, tstop, tstep)
+        times = np.arange(tmin, tmax + sys.float_info.epsilon, tstep)
+        if times[-1] > tmax:
```
christianbrodbeck: replaced with if...

agramfort: now I am happy :)

agramfort:
let's see if Travis and @Eric89GXL are happy too

surfer/viz.py
```diff
@@ -2178,19 +2179,16 @@ def save_movie(self, fname, time_dilation=4., tmin=None, tmax=None,
...
                              "(%r)" % (tmax, self._times[-1]))
-
-        # find indexes at which to create frames
-        tstep = 1. / (framerate * time_dilation)
-        if np.allclose((tmax - tmin) % tstep, 0):
-            tstop = tmax + tstep / 2.
-        else:
-            tstop = tmax
-        times = np.arange(tmin, tstop, tstep)
+        times = np.arange(tmin, tmax + sys.float_info.epsilon, tstep)
```
Eric89GXL: I don't think this guarantees what you think it does... epsilon should be the delta relative to 1, so if times steps are e.g. in the thousands this will not work. A cleaner way to do it is probably to do `np.arange(n_pts) * tstep`, then you're guaranteed the proper length.

Eric89GXL: e.g.
```
>>> 1. + sys.float_info.epsilon == 1.
False
>>> 2. + sys.float_info.epsilon == 2.
True
```

surfer/viz.py
```diff
@@ -2178,19 +2179,16 @@ def save_movie(self, fname, time_dilation=4., tmin=None, tmax=None,
...
-        tstep = 1. / (framerate * time_dilation)
-        if np.allclose((tmax - tmin) % tstep, 0):
-            tstop = tmax + tstep / 2.
-        else:
-            tstop = tmax
-        times = np.arange(tmin, tstop, tstep)
+        n_frames = floor((tmax - tmin) * time_dilation * framerate)
+        times = np.arange(n_frames)
+        times /= framerate * time_dilation
+        times += tmin
```
christianbrodbeck: Thanks @Eric89GXL , this seems to work for me.

Eric89GXL:
Thanks for another fix @christianbrodbeck

PR #135, opened 2015-07-05
DOC: change location of mailing list

matthew-brett:
Point to new neuroimaging mailing list


Eric89GXL:
Thanks @matthew-brett

PR #139, opened 2015-08-28
Allow brain without x offset

aestrivex:
The automatic offset creates a problem with my program which uses the vertex coordinates for a complicated algorithm to snap subcortical locations to optimally selected surface vertices and then plots the results. Easiest fix is to allow the user to disable this feature.

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -318,7 +318,7 @@ class Brain(object):
     views : list | str
         views to use
     disable_offset : bool
-        If True, disables alignment of origin with medial wall, instead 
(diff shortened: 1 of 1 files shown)


mwaskom:
I think I'd prefer this as `offset=True` rather than `disable_offset=False`, do others have thoughts?

aestrivex:
Doesn't matter to me, I implemented the change.

Eric89GXL:
LGTM

mwaskom:
Same here, thanks!

PR #145, opened 2015-11-02
Ffmpeg

christianbrodbeck:
Two small issues:

- ffmpeg says “Please use -b:a or -b:v, -b is ambiguous”
- Even with this, ffmpeg seems to print to stderr instead of stdout (on my setup), so logging an error seems wrong


agramfort:
shall I merge?

christianbrodbeck:
As far as I'm concerned it's ready, unless someone wants to verify whether FFmpeg also writes to stderr on other platforms

Eric89GXL:
Thanks @christianbrodbeck

PR #151, opened 2016-03-02
fix import Image in make_montage

### LINKED ISSUE #150: make_montage error
Hi all, I get the following error when I try to use make_montage or save_montage.

```
File "/anaconda/lib/python2.7/site-packages/surfer/viz.py", line 73, in make_montage
    import Image
ImportError: No module named Image
```

In ipython, I can do "from PIL import Image" just fine but just "import Image" by itself gives the same error. I tried to look at the link mentioned in the comments just prior to this line in make_montage but it did not illuminate anything for me... Am I missing some library, or is there something else I need to do to make this work?

Thank you!

natalieklein:
changed import line in make_montage to resolve #150

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -73,7 +73,7 @@ def make_montage(filename, fnames, orientation='h', colorbar=None,
     out : array
         The montage image data array.
     """
-    import Image
+    from PIL import Image
     # This line is only necessary to overcome a PIL bug, see:
     #     http://stackoverflow.com/questions/10854903/what-is-causing-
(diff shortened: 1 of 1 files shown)


mwaskom:
I don't really understand all the issues with PIL, but for the sake of backwards compatibility (in case some people can import `Image` but not `PIL` might it make sense to check for both with a try statement?

natalieklein:
@mwaskom Thanks, I don't really understand the issue either, but I hope this will work for everyone.

agramfort:
thx @natalieklein

PR #152, opened 2016-05-16
MRG: Py3k support

### LINKED ISSUE #143: Python 3 support
Just wanted to note that it looks like progress is being made in getting Mayavi on Python3: https://github.com/enthought/mayavi/pull/250

We should keep an eye on that and try to get PySurfer working when Mayavi stabilizes.

Eric89GXL:
Brings PySurfer py3k support. No CI support yet because Anaconda packages don't exist yet. But tests pass on my system with Mayavi dev and VTK 7.0.

Closes #143.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,9 +6,11 @@
 #                         Martin Luessi
 #                         Eric Larson
 
+import os
+from setuptools import setup
+
 descr = """PySurfer: cortical surface visualization using Python."""
 
-import os
 # deal with MPL sandbox violations during easy_install
 os.environ['MPLCONFIGDIR'] = '.'
 
@@ -32,8 +34,8 @@ LICENSE = 'BSD (3-clause)'
 DOWNLOAD_URL = 'https://github.com/nipy/PySurfer'
 VERSION = version
 
-def check_dependencies():
 
+def check_dependencies():
     needed_deps = ["IPython",
                    "numpy", "scipy", "matplotlib",
                    "mayavi",
@@ -50,8 +52,6 @@ def check_dependencies():
         raise ImportError("Missing dependencies: %s" % missing)
 
 
-from setuptools import setup
-
 if __name__ == "__main__":
     if os.path.exists('MANIFEST'):
         os.remove('MANIFEST')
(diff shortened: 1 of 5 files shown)


agramfort:
Can you build the mne doc with it?

Eric89GXL:
I have some annoying backend conflicts with `matplotlib` when I try to do it, but I don't think it's PySurfer's problem

setup.py
```diff
@@ -32,10 +34,9 @@
 DOWNLOAD_URL = 'https://github.com/nipy/PySurfer'
 VERSION = version
 
-def check_dependencies():
 
-    needed_deps = ["IPython",
```
Eric89GXL: I don't know why we had `IPython` listed as a dependency... @mwaskom looks like this was your change any idea?

agramfort:
LGTM

setup.py
mwaskom: The `pysurfer` command line tool drops you into IPython. Also in general you need IPython to do any interactive work.

Eric89GXL: > The pysurfer command line tool drops you into IPython.

Then it should be an optional dependency, not a hard requirement. People (like me) who do `import surfer` and do everything from there shouldn't need to have IPython.

> Also in general you need IPython to do any interactive work.

I disagree -- I use `python -i` (effectively) from Spyder in general, and rarely (if ever) use IPython, and have no problems with interactivity.

mwaskom: Ok. I don't particularly care about declaring it as a dependency...

Eric89GXL: Cool. Any other comments? If not I think it should be good to merge.

mwaskom: LGTM!

Eric89GXL:
Thanks for the quick reviews. Now everyone gets to have fun building the VTK and Mayavi :)

mwaskom:
Forgot to say cheers for doing the boring work here!

(But also, sotto voce boos for removing my main excuse for not switching to Python 3).

agramfort:
no pain installing mayavi on your mac michael?

mwaskom:
That's my current supplementary excuse. I haven't actually tried to use it yet :)

PR #154, opened 2016-06-26
[WIP] Imageio

christianbrodbeck:
Picking up an old suggestion from @Eric89GXL this uses [imageio](http://imageio.readthedocs.io/en/latest/index.html) for saving movies. Imageio downlaods and manages the FFMPEG binary automatically so this would provide considerable simplification.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,8 +29,7 @@ install:
         travis_retry sudo apt-get install mencoder;
       fi;
     - pip install -q coverage coveralls nose-timer nibabel flake8
-    - python setup.py build
-    - python setup.py install
+    - pip install .[save_movie]
     - SRC_DIR=$(pwd)
 
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,8 +28,9 @@ install:
         travis_retry sudo apt-get update -qq;
         travis_retry sudo apt-get install mencoder;
       fi;
-    - pip install -q coverage coveralls nose-timer nibabel flake8
-    - pip install .[save_movie]
+    - pip install -q coverage coveralls nose-timer nibabel flake8 imageio
+    - python setup.py build
+    - python setup.py install
     - SRC_DIR=$(pwd)
 
 before_script:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,8 +23,8 @@ install:
 
   # Install the dependencies of the project.
   - "conda install --yes --quiet ipython==1.1.0 numpy scipy mayavi matplotlib nose imaging"
-  - "pip install -q nose-timer"
-  - "pip install -e .[save_movie]"
+  - "pip install -q nose-timer nibabel imageio"
+  - "python setup.py develop"
   - "SET SUBJECTS_DIR=%CD%\\subjects"
 
 build: false  # Not a C# project, build stuff at the test step instead.
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,8 +23,8 @@ install:
 
   # Install the dependencies of the project.
   - "conda install --yes --quiet ipython==1.1.0 numpy scipy mayavi matplotlib nose imaging"
-  - "pip install -q nose-timer nibabel"
-  - "python setup.py develop"
+  - "pip install -q nose-timer"
+  - "pip install -e .[save_movie]"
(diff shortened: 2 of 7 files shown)


surfer/tests/test_viz.py
```diff
@@ -1,18 +1,18 @@
...
 import os.path as op
 from os.path import join as pjoin
-import re
 import shutil
-import subprocess
-from nose.tools import assert_equal
-from numpy.testing import assert_raises, assert_array_equal
 from tempfile import mkdtemp, mktemp
+
+import imageio
```
agramfort: not sure we want a strong dependency here.

agramfort:
nice to see these red lines without reducing features :)

surfer/tests/test_viz.py
christianbrodbeck: I added it as a dependency to `setup.py` to make usage seamless (it's on PYPI). If we remove that I assume we would also need some instructions about installing the dependency...

agramfort: ok but I would still make it optional. It's stupid to loose users for a use
case that represents a fraction of them. I don't have experience with this
lib to tell you if it actually works smoothly for everyone.

christianbrodbeck:
@agramfort I turned it into an optional dependency

agramfort:
@mwaskom @Eric89GXL any objection?

surfer/viz.py
```diff
@@ -2175,11 +2196,18 @@ def save_movie(self, fname, time_dilation=4., tmin=None, tmax=None,
...
+            $ pip install -U pysurfer[save_movie]
         """
-        assert_ffmpeg_is_available()
+        try:
+            import imageio
+        except ImportError:
+            raise ImportError("Saving movies from PySurfer requires the "
+                              "imageio library. To install imageio with "
+                              "PySurfer, run\n\n    $ pip install -U "
+                              "pysurfer[save_movie]\n")
```
Eric89GXL: Why not just suggest `pip install --upgrade imageio` directly? The suggestion will do more than just satisfy the ImageIO requirement, it will also potentially upgrade PySurfer, which I don't think is the cleanest suggestion.

Eric89GXL:
Otherwise LGTM

surfer/viz.py
christianbrodbeck: Right, I updated the message

.travis.yml
```diff
@@ -29,8 +29,7 @@ install:
         travis_retry sudo apt-get install mencoder;
       fi;
     - pip install -q coverage coveralls nose-timer nibabel flake8
-    - python setup.py build
-    - python setup.py install
+    - pip install .[save_movie]
```
agramfort: @Eric89GXL are you fine with this? It's the first I see this syntax

Eric89GXL: I assume the `.` means "the local package", and the brackets are optional deps. I haven't seen it either, but the CIs pass

christianbrodbeck: Yes, the alternative would be to add imageio to the `pip install` command, but this way we test at the same time that the optional dependency is resolved

agramfort: that would make the change more natural to me.

christianbrodbeck: Do you prefer me to revert?

agramfort: yes please.

appveyor.yml
```diff
@@ -23,7 +23,7 @@ install:
 
   # Install the dependencies of the project.
   - "conda install --yes --quiet ipython==1.1.0 numpy scipy mayavi matplotlib nose imaging"
-  - "pip install -q nose-timer nibabel"
+  - "pip install -q nose-timer nibabel imageio"
```
christianbrodbeck: reverted @agramfort

mwaskom:
LGTM

christianbrodbeck:
I might make sense to update the save_movie parameters to give better access to the imageio options (http://imageio.readthedocs.io/en/latest/format_ffmpeg.html#parameters-for-saving) but that could be a second PR

Eric89GXL:
Yeah a second PR would be good. Thanks @christianbrodbeck

christianbrodbeck:
Thanks!

PR #155, opened 2016-06-28
WIP:  adapt Brain.save_movie() API for imageio

christianbrodbeck:
Following up on https://github.com/nipy/PySurfer/pull/154

`imageio` parameters: http://imageio.readthedocs.io/en/latest/format_ffmpeg.html#parameters-for-saving

Probably the most sense would make to provide access to all of `imageio`'s options. I've saved a couple of movies and `imageio` seems to have sensible defaults (variable bitrate, hence I removed our fixed bitrate default). We could keep the existing keyword arguments so we don't break code that used positional arguments. Are there better alternatives? I am not sure how to document in case we just include a `**kwargs` in the signature. Should the documentation just include a link to `imageio`'s documentation?

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -2161,8 +2161,8 @@ class Brain(object):
         return out
 
     def save_movie(self, fname, time_dilation=4., tmin=None, tmax=None,
-                   framerate=24, interpolation='quadratic', codec='mpeg4',
-                   bitrate='1M'):
+                   framerate=24, interpolation='quadratic', codec=None,
+                   bitrate=None, **kwargs):
         """Save a movie (for data with a time axis)
 
         .. Warning::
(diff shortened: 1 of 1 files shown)


surfer/viz.py
```diff
@@ -2188,11 +2188,6 @@ def save_movie(self, fname, time_dilation=4., tmin=None, tmax=None,
             Interpolation method (``scipy.interpolate.interp1d`` parameter,
             one of 'linear' | 'nearest' | 'zero' | 'slinear' | 'quadratic' |
             'cubic', default 'quadratic').
-        codec : str
-            Codec to use with ffmpeg (default 'mpeg4').
-        bitrate : str | float
-            Bitrate to use to encode movie. Can be specified as number (e.g.
-            64000) or string (e.g. '64k'). Default value is 1M
```
agramfort: add kwargs to the docstring

christianbrodbeck:
Updated. Keeping the specific keywords open would have the big advantage that additional imageio formats would become available, such as *.gif

surfer/viz.py
```diff
@@ -2188,11 +2196,8 @@ def save_movie(self, fname, time_dilation=4., tmin=None, tmax=None,
...
+        additional keywords :
```
agramfort: why additional keywords and not `**kwargs : `

christianbrodbeck: fine with me :)

agramfort:
ready to merge?

christianbrodbeck:
If you're happy with the documentation

agramfort:
ok

thx @christianbrodbeck 

next time set the PR to MRG so I don't have to ask :)

christianbrodbeck:
Ok, sorry :)

PR #159, opened 2016-07-09
NF: added the option to specify alpha transparency for Brain binarized curvature rendering

ctw:
I have added the option to specify alpha transparency for Brain binarized curvature rendering. This can be useful when plotting foci in internal structures that are otherwise obscured by the brain surface. I had considered adding an "alpha" keyword to the Brain interface, but opted instead to allow for alpha transparency to be specified in the cortex keyword. I've implemented the chance in a way that should ensure backward compatibility with code specifying the cortex keyword as a 4-tuple or string (in these cases alpha defaults to 1.0). Using a new alpha transparency keyword for the Brain constructor might make this feature more accessible and I'm happy to change the implementation accordingly if that's preferred.

### DIFF
--- a/bin/pysurfer
+++ b/bin/pysurfer
@@ -73,9 +73,10 @@ if __name__ == '__main__':
         from surfer import Brain
 
         # Load  up the figure and underlying brain object
-        b = Brain(args.subject_id, args.hemi, args.surf, args.curv,
-                  args.title, views=args.views, size=args.size,
-                  background=args.background, cortex=args.cortex)
+        b = Brain(args.subject_id, args.hemi, args.surf, title=args.title,
+                  cortex=args.cortex, alpha=args.alpha, size=args.size,
+                  background=args.background, foreground=args.foreground,
+                  views=args.views)
 
         # Maybe load some morphometry
         if args.morphometry is not None:
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -566,12 +566,12 @@ class Brain(object):
                                        vmin=-.2, vmax=2,
                                        opacity=alpha), True, True))
         if isinstance(cortex, dict):
-            if not 'opacity' in cortex:
+            if 'opacity' not in cortex:
                 cortex['opacity'] = alpha
             if 'colormap' in cortex:
-                if not 'vmin' in cortex:
+                if 'vmin' not in cortex:
                     cortex['vmin'] = -1
-                if not 'vmax' in cortex:
+                if 'vmax' not in cortex:
                     cortex['vmax'] = 2
             geo_params = cortex, False, True
         elif cortex in colormap_map:
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -604,7 +604,7 @@ class Brain(object):
             geo_params = dict(colormap=cortex[0], vmin=cortex[1],
                               vmax=cortex[2], opacity=alpha), cortex[3], True
         else:
-            try: # check if it's a non-string color specification
+            try:  # check if it's a non-string color specification
                 color = colorConverter.to_rgb(cortex)
                 geo_params = dict(color=color, opacity=alpha), False, False
             except ValueError:
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -297,21 +297,24 @@ class Brain(object):
     title : str
         title for the window
     cortex : str, tuple, dict, or None
-        Specifies how the cortical surface is rendered. Accepts
-        specification of a colormap (in which case binarized curvature
-        values are rendered) or a color (in which case the cortical
-        surface is rendered without binarized curvature values).  Can
-        be set to: (1) the name of one of the preset cortex styles
-        ('classic' [default], 'high_contrast', 'low_contrast', or
-        'bone'), (2) the name of a colormap, (3) a tuple with four
-        entries for (colormap, vmin, vmax, reverse) indicating the
-        name of the colormap, the min and max values respectively and
-        whether or not the colormap should be reversed, (4) a valid
-        color specification (such as a 3-tuple with RGB values or a
-        valid color name), or (5) a dictionary of keyword arguments
-        that is passed on to the call to surface. If set to `None`,
-        color is set to (0.5, 0.5, 0.5) rendering a gray brain without
-        binarized curvature.
+        Specifies how the cortical surface is rendered. Options:
+
+            1. The name of one of the preset cortex styles:
+               ``'classic'`` (default), ``'high_contrast'``,
+               ``'low_contrast'``, or ``'bone'``.
+            2. A color-like argument to render the cortex as a single
+               color, e.g. ``'red'`` or ``(0.1, 0.4, 1.)``. Setting
+               this to ``None`` is equivalent to ``(0.5, 0.5, 0.5)``.
+            3. The name of a colormap used to render binarized
+               curvature values, e.g., ``Grays``.
+            4. A container with four entries for colormap (string
+               specifiying the name of a colormap), vmin (float
+               specifying the minimum value for the colormap), vmax
+               (float specifying the maximum value for the colormap),
+               and reverse (bool specifying whether the colormap
+               should be reversed. E.g., ``('Greys', -1, 2, False)``.
+            5. A dict of keyword arguments that is passed on to the
+               call to surface.
     alpha : float in [0, 1]
         Alpha level to control opacity of the cortical surface.
     size : float or pair of floats
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -296,20 +296,24 @@ class Brain(object):
         freesurfer surface mesh name (ie 'white', 'inflated', etc.)
     title : str
         title for the window
-    cortex : str, tuple, or dict
-        Specifies if/how binarized curvature values are rendered.  If
-        binarized curvature values should be rendered it can be
-        either the name of a preset PySurfer cortex colorscheme (one
-        of 'classic' (default), 'bone', 'low_contrast', or
-        'high_contrast'), the name of a mayavi colormap, a 4-tuple with
-        values (colormap, min, max, reverse), or a dictionary with
-        corresponding keys to fully specify the curvature colors. If a
-        color is specified instead, the brain is rendered in that
-        color without displaying the binarized curvature values. A
-        setting of `None` produces a gray brain without binarized
-        curvature.
+    cortex : str, tuple, dict, or None
+        Specifies how the cortical surface is rendered. Accepts
+        specification of a colormap (in which case binarized curvature
+        values are rendered) or a color (in which case the cortical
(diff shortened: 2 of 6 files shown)


mwaskom:
> Using a new alpha transparency keyword for the Brain constructor might make this feature more accessible and I'm happy to change the implementation accordingly if that's preferred.

I think this would be my preference, but others can comment. It's also possible that `opacity` would be a more obvious parameter name.

mwaskom:
Also one weird thing about implementing it through the curvature overlay is that I would think in most cases when you want a semi-transparent cortical mesh you don't want to be showing curvature (it will only complicate things).

ctw:
I agree that it's a bit annoying that one can only specify alpha when curv is True in the current implementation (although by specifying sufficiently large absolute values for vmin and vmax one can effectively eliminate the contrast). Will change to use a keyword argument unless others object. It seems like "alpha" is the standard keyword argument for transparency (e.g., in `add_data`, `add_annotation`, `add_label`, and `add_foci`), that's why I chose it over `opacity`

Eric89GXL:
+1 for keyword

ctw:
On further reflection, I'd like to refactor the interface to Brain a bit more as follows:

I would like to deprecate the `curv` keyword in favor of allowing the cortex keyword to accept colors (in addition to the currently allowed inputs). If a color is specified, a solid brain (without the curvature) is drawn in that color. This would add the new functionality of allowing the user to specify the color a solid brain is drawn in and eliminate one keyword.

To summarize I propose the following changes:

- add `alpha` keyword to specify transparency of the brain surface.
- keep default value and behavior for currently accepted inputs to `cortex` as is, but also allow specification of a color (rather than color map) to produce a solid brain surface (without curvature) in that color. If `cortex` is set to `None` the same output is produced as if `curv` is set to `False` in the current interface.
- keep `curv` for backward compatibility, but set the default to `None` and throw a warning that it is deprecated and will be removed in a future version if it is changed from the default value.

Any thoughts?

agramfort:
no objection

ctw:
I've made the proposed changes. Please let me know if you have questions and/or suggestions for improvements.

surfer/viz.py
```diff
@@ -294,17 +294,22 @@ class Brain(object):
...
+        of 'classic' (default), 'bone', 'low_contrast', or
+        'high_contrast'), the name of a mayavi colormap, a 4-tuple with
+        values (colormap, min, max, reverse), or a dictionary with
+        corresponding keys to fully specify the curvature colors. If a
+        color is specified instead, the brain is rendered in that
+        color without displaying the binarized curvature values. A
+        setting of `None` produces a gray brain without binarized
+        curvature.
+    alpha : float in [0, 1]
+        Alpha level to control opacity.
```
mwaskom: I'd make this a little bit more specific about what opacity is being controlled.

surfer/viz.py
```diff
@@ -356,6 +361,18 @@ def __init__(self, subject_id, hemi, surf, curv=True, title=None,
...
+        # Keep backwards compatability
+        if curv is not None:
+            msg = ("The `curv` keyword has been deprecated and will "
+                   "be removed in future versions. You should update your "
+                   "code and use the `cortex` keyword to specify how the "
+                   "brain surface is rendered. Setting `cortex` to `None` "
+                   "will reproduce the previous behavior when `curv` was "
+                   "set to `False`. To emulate the previous behavior for "
+                   "cases where `curv` was set to `True`, simply omit it.")
+            warn(msg)
```
mwaskom: Very informative message. Should this have a specific warning class?

surfer/viz.py
```diff
@@ -492,6 +510,77 @@ def _set_window_properties(self, size, background, foreground):
...
 
+    def _get_geo_params(self, cortex, alpha=1.0):
+        """Return keyword arguments and other parameters for surface
+        rendering.
+
+        Parameters
+        ----------
+        cortex : {classic, high_contrast, low_contrast, bone, tuple}
+            The name of one of the preset cortex styles, or a tuple
+            with four entries as described in the return vales.
```
mwaskom: Looking at the return values documentation, it's not clear what the tuple entries should be. (There are three return values for this function).

surfer/viz.py
```diff
@@ -294,17 +294,22 @@ class Brain(object):
...
+        binarized curvature values should be rendered it can be
+        either the name of a preset PySurfer cortex colorscheme (one
+        of 'classic' (default), 'bone', 'low_contrast', or
+        'high_contrast'), the name of a mayavi colormap, a 4-tuple with
+        values (colormap, min, max, reverse), or a dictionary with
+        corresponding keys to fully specify the curvature colors. If a
+        color is specified instead, the brain is rendered in that
+        color without displaying the binarized curvature values. A
+        setting of `None` produces a gray brain without binarized
+        curvature.
```
mwaskom: I'm a little concerned about how many different options (all with different types) this parameter has. Just reading the keyword argument documentation, I feel a little overwhelmed. I'm not sure what the best solution is, but I think it would be good to think a little bit more about whether this logic can be made a bit simpler.

mwaskom:
Please make sure that you update the `pysurfer` script to reflect this changes and also that you add a note to the `CHANGES` file.

You will also want to look at the example scripts and make sure that none of them are using `curv`. Then you should probably add an example script to show a semi-transparent brain (perhaps demonstrating how to add subcortical foci). It would also be good to update at least one other example script to cover the new functionality for specifying the curvature values.

surfer/viz.py
ctw: Done!

surfer/viz.py
ctw: I've added the category DeprecationWarning to warnings for the use of config_opts and curv. Note that the default behavior for DeprecationWarning is that it's ignored, which makes these warnings somewhat less visible now.

surfer/viz.py
ctw: I've updated the docstring to make this clear.

surfer/viz.py
ctw: I've updated the docstring for clarity. I tried to highlight that it's basically just two high-level options (either a colormap or a color) and that there are different ways either can be specified.

ctw:
I've made the suggested changes. None of the existing example scripts were affected by the changes, but I've added a new one where alpha transparency is specified and a color is passed to cortex instead of a colormap.

mwaskom:
Thanks! Would you mind showing the output of the example script here in the thread so I don't need to pull down your changes and build the docs?

surfer/viz.py
mwaskom: You could use a `UserWarning` if you want it to be more visible.

surfer/viz.py
mwaskom: Also note the typo in this line.

bin/pysurfer
```diff
@@ -73,9 +73,10 @@ if __name__ == '__main__':
         from surfer import Brain
 
         # Load  up the figure and underlying brain object
-        b = Brain(args.subject_id, args.hemi, args.surf, args.curv,
-                  args.title, views=args.views, size=args.size,
-                  background=args.background, cortex=args.cortex)
+        b = Brain(args.subject_id, args.hemi, args.surf, title=args.title,
+                  cortex=args.cortex, alpha=args.alpha, size=args.size,
+                  background=args.background, foreground=args.foreground,
+                  views=args.views)
```
mwaskom: I don't think this change fully handles the case where someone supplies `-no-curv` on the command line.

ctw: You're right -- it should work now.

ctw:
I'm having trouble building the docs, at least partly, it seems, because I don't have freesurfer installed. Is there another way?

I like the option of being able to pass two colors to cortex, but the subsequent call to mlab.pipeline.surface expects the colormap to be specified by a string, so getting this to work would seem to require some refactoring.

mwaskom:
> I like the option of being able to pass two colors to cortex, but the subsequent call to mlab.pipeline.surface expects the colormap to be specified by a string, so getting this to work would seem to require some refactoring.

I think you can take a look in `Brain.add_data` to see how custom colormaps are set there.

surfer/viz.py
```diff
@@ -294,17 +294,26 @@ class Brain(object):
...
+        ('classic' [default], 'high_contrast', 'low_contrast', or
+        'bone'), (2) the name of a colormap, (3) a tuple with four
+        entries for (colormap, vmin, vmax, reverse) indicating the
+        name of the colormap, the min and max values respectively and
+        whether or not the colormap should be reversed, (4) a valid
+        color specification (such as a 3-tuple with RGB values or a
+        valid color name), or (5) a dictionary of keyword arguments
+        that is passed on to the call to surface. If set to `None`,
+        color is set to (0.5, 0.5, 0.5) rendering a gray brain without
+        binarized curvature.
```
Eric89GXL: Can you make this prettier (and easier to read) with proper ReStructuredText formatting like:
```
cortex : ...
    Specifies how the cortical surface is rendered. Options:

        1. A color-like argument to render the cortex as a single
           color, e.g. ``'red'`` or ``(0.1, 0.4, 1.)``.
        2. A colormap, in which binarized curvature ...
        3. ...
        ...
```
Why even allow `None`, isn't it the same as passing `(0.5, 0.5, 0.5)`?

Eric89GXL:
Otherwise the arguments look reasonable to me

surfer/viz.py
ctw: Done. I think it's nice to have a 'None' option to render a default gray brain. There are different ways to generate other colors as well (e.g., by passing in the name or the corresponding RGB values).

ctw:
I've added the option to pass a list of colors as argument for cortex as suggested by @mwaskom.

Eric89GXL:
Style:

https://travis-ci.org/nipy/PySurfer/jobs/144918615#L826

ctw:
Thanks @Eric89GXL -- fixed now!

Eric89GXL:
Looks like there are a number of [un-covered relevant lines] (https://coveralls.io/builds/7054113/source?filename=surfer%2Fviz.py#L575), can you write some simple tests that at least touch them? Can't really test the output but it can at least make sure some inputs get parsed.

Eric89GXL:
Otherwise LGTM, +1 for merge. Example looks cool.

Eric89GXL:
Ahh good I see "small fix" in the commit message to justify my nitpick :)

Eric89GXL:
PEP8 line length

https://travis-ci.org/nipy/PySurfer/jobs/145638217#L826

Eric89GXL:
@mwaskom feel free to merge if you're happy, too

ctw:
@mwaskom any chance this can be merged?

mwaskom:
Hi, sorry. I started my postdoc last month and also just adopted a dog so I'm dropping things left and right. But if this looks good to @Eric89GXL it looks good to me.

PR #161, opened 2016-08-05
Decorate project_volume_data with verbose (closes #160)

### LINKED ISSUE #160: Does verbose work?
The `verbose` parameter, specifically in `project_volume_data`, doesn't seem to have any effect. e.g.:

```python
surf = project_volume_data(vol, hemi, reg,
                               smooth_fwhm=1, projarg=(.5, .5, 1),
                               verbose=verbose)
```

Prints (to `stderr`, based on the Jupyter css color) INFO level log information no matter whether I set `verbose` to `False` or `"DEBUG"`.

Do I not understand how to use it, or is this parameter being ignored?

mwaskom:
Closes #160

### DIFF
--- a/surfer/io.py
+++ b/surfer/io.py
@@ -7,10 +7,10 @@ import numpy as np
 import nibabel as nib
 from nibabel.spatialimages import ImageFileError
 
+from .utils import verbose
+
 import logging
 logger = logging.getLogger('surfer')
-
-from .utils import verbose
 
 
 def read_scalar_data(filepath):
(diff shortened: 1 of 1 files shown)


surfer/io.py
```diff
@@ -10,6 +10,8 @@
 import logging
 logger = logging.getLogger('surfer')
 
+from .utils import verbose
```
Eric89GXL: flake error, module level import goes above calls. Move it up three or more lines

mwaskom:
Right right...it also offends my sensibilities to have a local import above a stdlib one, but I won't move more things around :)

PR #162, opened 2016-08-15
FIX : py3k viz support now that mayavi seems to actually work with py3

agramfort:


Eric89GXL:
Thanks @agramfort, I take it you're trying it now?

PR #179, opened 2017-02-08
FIX: Fix Travis

Eric89GXL:
Let's see if this works. From:

https://github.com/ContinuumIO/anaconda-issues/issues/1365

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,41 +1,40 @@
-language: python
-
+language: c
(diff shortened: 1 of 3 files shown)


Eric89GXL:
Looks like we're green! I'll merge so that other PRs can actually get feedback, but if people feel like looking and commenting feel free and I can fix anything in a follow-up PR

PR #185, opened 2017-03-13
[MRG] FIX Brain.remove labels()

christianbrodbeck:
Remove the `hemi` parameter, which was previously unused (even though it forced the user to set it). Labels are not stored by hemisphere, but by their name.


Eric89GXL:
LGTM. @mwaskom can you look?

mwaskom:
Code looks good to me. Are we still ignoring AppVeyor failures?

Eric89GXL:
It seems unrelated

PR #189, opened 2017-04-24
Use different interface for setting LUT

### LINKED ISSUE #188: System-dependent handling of saturated data values
I am getting pysurfer set up on our new Ubuntu system and have discovered something unusual. Values that are plotted with `Brain.add_data` are transparent when they are below the `min` or above the `max` values. i.e. they are being treated as if they are below the `threshold`, even when I am not using a threshold. This behavior does not occur on my OSX system, using github master on both machines. It also does not occur when using `Brain.add_overlay`.

The behavior can be reproduced using the `examples/plot_resting_correlations.py` script. Values around the "seed" focus are clipped/transparent in the final plot (but not in the intermediate plot created with `add_overlay`).

Can anyone else reproduce this? Does anyone have any ideas what the issue is?

mwaskom:
Closes #188

There is an inconsistency where the previous method wanted values in [0, 255] while the current method wants values in [0, 1]. I decided to handle this within each method to otherwise maintain prior API.

I generated the docs on my Ubuntu system and everything looks good. It would  be best to have someone else on a different system verify that the example gallery does not change with this PR.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,1 +1,1 @@
-Michael Waskom <mwaskom@stanford.edu> mwaskom <mwaskom@stanford.edu> Michael Waskom <mwaskom@mit.edu>
+Michael Waskom <mwaskom@stanford.edu> mwaskom <mwaskom@stanford.edu> Michael Waskom <mwaskom@mit.edu> Michael Waskom <mwaskom@nyu.edu>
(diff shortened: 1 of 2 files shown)


agramfort:
fine with me

Eric89GXL:
Thanks @mwaskom

Issue #177, opened 2017-02-08
ENH: Speed up rendering by sharing data


larsoner:
We have a lot of data that could be shared (vertex locations), and we should do it:

http://docs.enthought.com/mayavi/mayavi/tips.html#sharing-the-same-data-between-scenes

christianbrodbeck:
I'm working on this at #191

PR #191, opened 2017-05-18
[MRG]: Set surface

christianbrodbeck:
I often find myself plotting with one surface and then wanting to look at it with another. This implementation is based on keeping track of all the meshes that are added to the brain. I've implemented it for data, labels and annotations, not yet for all other layer types because I wanted to confirm first that there is no better way to implement it. I was wondering whether all the mesh objects could be combined into one somehow (re #177), but it looks like they all need to have a different `scalars` attribute, and I don't see a straightforward way to separate the scalars from the triangular mesh...

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,12 @@
 PySurfer Changes
 ================
 
+Version 0.8
+-----------
+
+- The surface geometry that is displayed can now be changed after initializing
+  a ``Brain`` instance with e.g. ``Brain.set_surf("smoothwm")``.
+
 Version 0.7
 -----------
 
--- a/CHANGES
+++ b/CHANGES
@@ -5,7 +5,7 @@ Version 0.8
 -----------
 
 - The surface geometry that is displayed can now be changed after initializing
-  a ``Brain`` instance with e.g. ``Brain.set_surf("smoothwm")``.
+  a ``Brain`` instance with e.g. ``brain.set_surf("smoothwm")``.
 
 Version 0.7
 -----------
--- a/surfer/tests/test_viz.py
+++ b/surfer/tests/test_viz.py
@@ -308,7 +308,7 @@ def test_movie():
 @requires_fsaverage
 def test_overlay():
     """Test plotting of overlay."""
-    mlab.options.backend = 'test'
+    mlab.options.backend = 'auto'
     # basic overlay support
     overlay_file = pjoin(data_dir, "lh.sig.nii.gz")
     brain = Brain(*std_args)
--- a/surfer/tests/test_viz.py
+++ b/surfer/tests/test_viz.py
@@ -100,6 +100,7 @@ def test_brains():
         brain = Brain(subject_id, hemi, surf, title=title, cortex=cort,
                       alpha=alpha, size=s, background=bg, foreground=fg,
                       figure=fig, subjects_dir=sd)
+        brain.set_distance()
         brain.close()
     brain = Brain(subject_id, hemi, surf, subjects_dir=sd,
                   interaction='terrain')
--- a/surfer/tests/test_viz.py
+++ b/surfer/tests/test_viz.py
@@ -120,7 +120,7 @@ def test_annot():
     brain = Brain(*std_args)
     for a, b, p in zip(annots, borders, alphas):
         brain.add_annotation(a, b, p)
-    brain.set_surface('smoothwm')
+    brain.set_surface('white')
     assert_raises(ValueError, brain.add_annotation, 'aparc', borders=-1)
     brain.close()
 
@@ -149,7 +149,7 @@ def test_data():
     reg_file = pjoin(data_dir, 'register.dat')
     surf_data = io.project_volume_data(mri_file, "lh", reg_file)
     brain.add_data(surf_data, -.7, .7, colormap="jet", alpha=.7)
-    brain.set_surface('smoothwm')
+    brain.set_surface('white')
     brain.add_data([], vertices=np.array([], int))
     brain.close()
 
@@ -193,7 +193,7 @@ def test_label():
     brain.add_label("V1", color="steelblue", alpha=.6)
     brain.add_label("V2", color="#FF6347", alpha=.6)
     brain.add_label("entorhinal", color=(.2, 1, .5), alpha=.6)
-    brain.set_surface('smoothwm')
+    brain.set_surface('white')
 
     # remove labels
     brain.remove_labels('V1')
@@ -248,7 +248,7 @@ def test_meg_inverse():
     assert_equal(data_dicts[1]['time_idx'], 0)
 
     # change surface
-    brain.set_surface('smoothwm')
+    brain.set_surface('white')
 
     # remove all layers
     brain.remove_data()
--- a/surfer/tests/test_viz.py
+++ b/surfer/tests/test_viz.py
@@ -357,8 +357,9 @@ def test_probabilistic_labels():
     prob_field[ids] = probs
     brain.add_data(prob_field, thresh=1e-5)
 
-    brain.data["colorbar"].number_of_colors = 10
-    brain.data["colorbar"].number_of_labels = 11
+    with warnings.catch_warnings(record=True):
+        brain.data["colorbar"].number_of_colors = 10
+        brain.data["colorbar"].number_of_labels = 11
     brain.close()
 
 
--- a/surfer/tests/test_viz.py
+++ b/surfer/tests/test_viz.py
@@ -113,7 +113,7 @@ def test_brains():
 @requires_fsaverage
 def test_annot():
     """Test plotting of annot."""
-    mlab.options.backend = 'test'
+    mlab.options.backend = 'auto'
     annots = ['aparc', 'aparc.a2005s']
     borders = [True, False, 2]
     alphas = [1, 0.5]
@@ -128,7 +128,7 @@ def test_annot():
 @requires_fsaverage
 def test_contour():
     """Test plotting of contour overlay."""
-    mlab.options.backend = 'test'
+    mlab.options.backend = 'auto'
     brain = Brain(*std_args)
     overlay_file = pjoin(data_dir, "lh.sig.nii.gz")
     brain.add_contour_overlay(overlay_file)
@@ -143,7 +143,7 @@ def test_contour():
 @requires_fs
 def test_data():
     """Test plotting of data."""
-    mlab.options.backend = 'test'
+    mlab.options.backend = 'auto'
     brain = Brain(*std_args)
     mri_file = pjoin(data_dir, 'resting_corr.nii.gz')
     reg_file = pjoin(data_dir, 'register.dat')
@@ -177,7 +177,7 @@ def test_foci():
 @requires_fsaverage
 def test_label():
     """Test plotting of label."""
-    mlab.options.backend = 'test'
+    mlab.options.backend = 'auto'
     subject_id = "fsaverage"
     hemi = "lh"
     surf = "inflated"
@@ -208,7 +208,7 @@ def test_label():
 @requires_fsaverage
 def test_meg_inverse():
     """Test plotting of MEG inverse solution."""
-    mlab.options.backend = 'test'
+    mlab.options.backend = 'auto'
     brain = Brain(*std_args)
     stc_fname = os.path.join(data_dir, 'meg_source_estimate-lh.stc')
     stc = io.read_stc(stc_fname)
@@ -260,7 +260,7 @@ def test_meg_inverse():
 @requires_fsaverage
 def test_morphometry():
     """Test plotting of morphometry."""
-    mlab.options.backend = 'test'
+    mlab.options.backend = 'auto'
     brain = Brain(*std_args)
     brain.add_morphometry("curv")
     brain.add_morphometry("sulc", grayscale=True)
@@ -340,7 +340,7 @@ def test_overlay():
 @requires_fsaverage
 def test_probabilistic_labels():
     """Test plotting of probabilistic labels."""
-    mlab.options.backend = 'test'
+    mlab.options.backend = 'auto'
     brain = Brain("fsaverage", "lh", "inflated",
                   cortex="low_contrast")
 
--- a/surfer/tests/test_viz.py
+++ b/surfer/tests/test_viz.py
@@ -120,6 +120,7 @@ def test_annot():
     brain = Brain(*std_args)
     for a, b, p in zip(annots, borders, alphas):
         brain.add_annotation(a, b, p)
+    brain.set_surface('smoothwm')
     assert_raises(ValueError, brain.add_annotation, 'aparc', borders=-1)
     brain.close()
 
@@ -148,6 +149,7 @@ def test_data():
     reg_file = pjoin(data_dir, 'register.dat')
     surf_data = io.project_volume_data(mri_file, "lh", reg_file)
     brain.add_data(surf_data, -.7, .7, colormap="jet", alpha=.7)
+    brain.set_surface('smoothwm')
     brain.add_data([], vertices=np.array([], int))
     brain.close()
 
@@ -191,6 +193,7 @@ def test_label():
     brain.add_label("V1", color="steelblue", alpha=.6)
     brain.add_label("V2", color="#FF6347", alpha=.6)
     brain.add_label("entorhinal", color=(.2, 1, .5), alpha=.6)
+    brain.set_surface('smoothwm')
 
     # remove labels
     brain.remove_labels('V1')
@@ -243,6 +246,9 @@ def test_meg_inverse():
     brain.set_data_time_index(0)
     assert_equal(data_dicts[0]['time_idx'], 0)
     assert_equal(data_dicts[1]['time_idx'], 0)
+
+    # change surface
+    brain.set_surface('smoothwm')
 
     # remove all layers
     brain.remove_data()
--- a/surfer/tests/test_viz.py
+++ b/surfer/tests/test_viz.py
@@ -334,6 +334,9 @@ def test_overlay():
     brain.add_overlay(conjunct, 4, 30, name="conjunct")
     brain.overlays["conjunct"].pos_bar.lut_mode = "Purples"
     brain.overlays["conjunct"].pos_bar.visible = False
+
+    brain.set_surface('white')
+
     brain.close()
 
 
--- a/surfer/tests/test_viz.py
+++ b/surfer/tests/test_viz.py
@@ -352,7 +352,7 @@ def test_probabilistic_labels():
     brain.add_label("BA45", color="salmon", borders=True, scalar_thresh=.5)
 
     label_file = pjoin(subj_dir, "fsaverage", "label", "lh.BA6.label")
-    prob_field = np.zeros_like(brain._geo.x)
+    prob_field = np.zeros_like(brain.geo['lh'].x)
     ids, probs = nib.freesurfer.read_label(label_file, read_scalars=True)
     prob_field[ids] = probs
     brain.add_data(prob_field, thresh=1e-5)
--- a/surfer/tests/test_viz.py
+++ b/surfer/tests/test_viz.py
@@ -120,7 +120,7 @@ def test_annot():
     brain = Brain(*std_args)
     for a, b, p in zip(annots, borders, alphas):
         brain.add_annotation(a, b, p)
-    brain.set_surface('white')
+    brain.set_surf('white')
     assert_raises(ValueError, brain.add_annotation, 'aparc', borders=-1)
     brain.close()
 
@@ -149,7 +149,7 @@ def test_data():
     reg_file = pjoin(data_dir, 'register.dat')
     surf_data = io.project_volume_data(mri_file, "lh", reg_file)
     brain.add_data(surf_data, -.7, .7, colormap="jet", alpha=.7)
-    brain.set_surface('white')
+    brain.set_surf('white')
     brain.add_data([], vertices=np.array([], int))
     brain.close()
 
@@ -193,7 +193,7 @@ def test_label():
     brain.add_label("V1", color="steelblue", alpha=.6)
     brain.add_label("V2", color="#FF6347", alpha=.6)
     brain.add_label("entorhinal", color=(.2, 1, .5), alpha=.6)
-    brain.set_surface('white')
+    brain.set_surf('white')
 
     # remove labels
     brain.remove_labels('V1')
@@ -248,7 +248,7 @@ def test_meg_inverse():
     assert_equal(data_dicts[1]['time_idx'], 0)
 
     # change surface
-    brain.set_surface('white')
+    brain.set_surf('white')
 
     # remove all layers
     brain.remove_data()
@@ -335,7 +335,7 @@ def test_overlay():
     brain.overlays["conjunct"].pos_bar.lut_mode = "Purples"
     brain.overlays["conjunct"].pos_bar.visible = False
 
-    brain.set_surface('white')
+    brain.set_surf('white')
 
     brain.close()
 
--- a/surfer/utils.py
+++ b/surfer/utils.py
@@ -32,7 +32,7 @@ class _MayaviThresholdPatch(object):
     In Mayavi 4.5, filters seem to be missing a .point_data attribute that
     Threshold accesses on initialization.
     """
-    need_patch = LooseVersion(mayavi.__version__) <= LooseVersion('4.5.0')
+    need_patch = LooseVersion(mayavi.__version__) == LooseVersion('4.5.0')
     _orig_func = Threshold._get_data_range
 
     @staticmethod
--- a/surfer/utils.py
+++ b/surfer/utils.py
@@ -34,7 +34,10 @@ class _MayaviThresholdPatch(object):
     """
     need_patch = LooseVersion(mayavi.__version__) <= LooseVersion('4.5.0')
     _orig_func = Threshold._get_data_range
-    _patch_func = lambda x: []
+
+    @staticmethod
+    def _patch_func(self):
+        return []
 
     def __enter__(self):
         if self.need_patch:
--- a/surfer/utils.py
+++ b/surfer/utils.py
@@ -8,6 +8,7 @@ import inspect
 from functools import wraps
 
 import mayavi
+from mayavi import mlab
 from mayavi.filters.api import Threshold
 import numpy as np
 import nibabel as nib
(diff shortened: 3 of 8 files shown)


mwaskom:
Cool!

My first thought is that it should probably be `set_surf` to be more consistent with the keyword in the `Brain` constructor (and the terminology in most of Freesurfer.

christianbrodbeck:
I'm working on sharing data based on [this](http://gael-varoquaux.info/programming/mayavi-representing-an-additional-scalar-on-surfaces.html). I ran into an error with Threshold filters which try to access a nonexistent `.point_data` attribute:

screen shot 2017-06-13 at 8 12 34 pm

Then I noticed a similar error happens when I I run the second (longer) example from the link. Is this a bug in Mayavi 4.5? Can others run the example without getting error dialogs?

I also had to set the backend for many tests to `auto` to avoid getting this dialog:

screen shot 2017-06-13 at 8 25 39 pm

and this error:

```
apptools.preferences.preferences: DEBUG: loading preferences from </Users/christian/.enthought/mayavi_e3/preferences.ini>
apptools.preferences.preferences: DEBUG: loading preferences from <<open file '//anaconda/envs/mayavi45/lib/python2.7/site-packages/mayavi/preferences/preferences.ini', mode 'rb' at 0x1188470c0>>
apptools.preferences.preferences: DEBUG: loading preferences from <<open file '//anaconda/envs/mayavi45/lib/python2.7/site-packages/tvtk/plugins/scene/preferences.ini', mode 'rb' at 0x1188478a0>>
mayavi.core.registry: DEBUG: Engine [<mayavi.core.null_engine.NullEngine object at 0x102d38e30>] named NullEngine1 registered
mayavi.core.common: ERROR: No valid current object, please select an active object.
traits: ERROR: Exception occurred in traits notification handler for object: <mayavi.tools.filters.SetActiveAttributeFactory object at 0x1029d8b30>, trait: point_scalars, old value: , new value: 2
Traceback (most recent call last):
  File "//anaconda/envs/mayavi45/lib/python2.7/site-packages/traits/trait_notifiers.py", line 340, in __call__
    self.handler( *args )
  File "//anaconda/envs/mayavi45/lib/python2.7/site-packages/mayavi/tools/pipe_base.py", line 198, in _anytrait_changed
    setattr(obj, components[-1], value)
  File "//anaconda/envs/mayavi45/lib/python2.7/site-packages/mayavi/core/trait_defs.py", line 97, in set_value
    raise TraitError(object, name,"one of %s"%values, value)
TraitError: The 'point_scalars_name' trait of a SetActiveAttribute instance must be one of [], but a value of '2' <type 'str'> was specified.
```

christianbrodbeck:
I've opened an issue at https://github.com/enthought/mayavi/issues/511

Eric89GXL:
I think 4.4.4 (or whatever it was) should still live in the Anaconda ecosystem somewhere, can you see if you can run it on older Mayavi version(s)?

That blog post is over 7 years old, so I'm not too surprised it no longer works :(

Eric89GXL:
On my OSX machine I get no errors. On my Linux machine, I get:
screenshot from 2017-06-14 10-44-40

Followed by:
screenshot from 2017-06-14 10-44-56

with messages:
```
No handlers could be found for logger "mayavi.core.common"
Exception occurred in traits notification handler.
Please check the log file for details.
Exception occurred in traits notification handler for object: <mayavi.tools.decorations.Axes object at 0x7f6c2b295c50>, trait: extent, old value: [0 0 0 0 0 0], new value: [ -50.   49.  -50.   49.    0.  100.]
Traceback (most recent call last):
  File "/home/larsoner/anaconda2/lib/python2.7/site-packages/traits/trait_notifiers.py", line 340, in __call__
    self.handler( *args )
  File "/home/larsoner/anaconda2/lib/python2.7/site-packages/mayavi/tools/decorations.py", line 373, in _extent_changed
    axes.module_manager.source.outputs[0].bounds
AttributeError: 'AssignAttribute' object has no attribute 'bounds'
```
But the plot works (here I've replaced `w = np.random.RandomState(0).randn(*z.shape)`):
screenshot from 2017-06-14 10-44-07

And *I get no errors if I remove the mlab.axes() call*. Can you see if it's the case for you?

christianbrodbeck:
Yes you're right, it was just the `mlab.axes()` call!

So now for the actual problem: If I replace the `surf = ...` line with 

```
thresh = mlab.pipeline.threshold(active_attr, low=0.5)
surf = mlab.pipeline.surface(thresh)
```

 I get:

screen shot 2017-06-14 at 11 07 25 am

then

screen shot 2017-06-14 at 11 07 28 am

and

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/Users/christian/Documents/scripts/mayavi/scalar_data.py in <module>()
     34 ###
     35 thresh = mlab.pipeline.threshold(active_attr, low=0.5)
---> 36 surf = mlab.pipeline.surface(thresh)
     37 ###
     38 

/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/mayavi/tools/pipe_base.pyc in the_function(*args, **kwargs)
     36 def make_function(factory_class):
     37     def the_function(*args, **kwargs):
---> 38         factory = factory_class(*args, **kwargs)
     39         return factory._target
     40 

/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/mayavi/tools/modules.pyc in __init__(self, *args, **kwargs)
    154 
    155     def __init__(self, *args, **kwargs):
--> 156         super(DataModuleFactory, self).__init__(*args, **kwargs)
    157         # We are adding data to the scene, reset the zoom:
    158         scene = self._scene.scene

/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/mayavi/tools/pipe_base.pyc in __init__(self, parent, **kwargs)
    161         # Now calling the traits setter, so that traits handlers are
    162         # called
--> 163         self.set(**traits)
    164         if scene is not None:
    165             scene.disable_render = not self._do_redraw

/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/mayavi/tools/pipe_base.pyc in set(self, trait_change_notify, **traits)
    177             try:
    178                 if callback is not None:
--> 179                     callback()
    180                 self._anytrait_changed(trait, value)
    181             except TraitError:

/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/mayavi/tools/modules.pyc in _colormap_changed(self)
    123             self._target.module_manager.scalar_lut_manager.reverse_lut = True
    124             self._target.module_manager.vector_lut_manager.reverse_lut = True
--> 125         self._target.module_manager.scalar_lut_manager.lut_mode = colormap
    126         self._target.module_manager.vector_lut_manager.lut_mode = colormap
    127 

AttributeError: 'NoneType' object has no attribute 'scalar_lut_manager'
```

Eric89GXL:
I see the same error. Can you confirm that this fixes it? Change `mayavi/filters/threshold.py` line ~223:
```
    def _get_data_range(self):
        """Returns the range of the input scalar data."""
        input = self.inputs[0].outputs[0]
        data_range = []
        if hasattr(input, 'point_data'):
            ps = input.point_data.scalars
        else:
            ps = None
        if hasattr(input, 'cell_data'):
            cs = input.cell_data.scalars
        else:
            cs = None
```

Eric89GXL:
(The Mayavi people have a ton of issues open, but you could open a PR as a potential fix, including some minimal example -- it makes it more likely for a proper fix to land)

christianbrodbeck:
With the dev version of Mayavi I don't get this error, so I assume it has been fixed already (adding the missing attributes to the other filters)...

So for this PR, I assume the best course is to build in a workaround with a comment about what can be changed once mayavi updates?

Eric89GXL:
> So for this PR, I assume the best course is to build in a workaround with a comment about what can be changed once mayavi updates?

Can we work around it by checking the mayavi version and monkey-patching a `None`?

surfer/viz.py
```diff
@@ -30,6 +32,12 @@
 logger = logging.getLogger('surfer')
 
 
+# Monkey-patch for Mayavi 4.5: filters are missing a point_data attribute that
+# Threshold accesses on initialization
+if LooseVersion(mayavi.__version__) <= LooseVersion('4.5.0'):
+    mayavi.filters.api.Threshold._get_data_range = lambda x: []
```
christianbrodbeck: This fixes it – maybe not ideal to modify the class but acceptable for this purpose since it does not seem to have side-effects?

Eric89GXL: Yeah it's heavy-handed but probably our only option.

One other thing we could do that would be less brutal is make a context manager that *temporarily* monkey-patches, and then sets back once the potentially offensive line completes, like:
```
with PatchedMayavi():
    surf = ...
```

christianbrodbeck:
While updating all the displays I noticed that overlays are handled in a way that is not quite consistent with other objects, by being implemented with two classes (one to represent data, one to display it) while all other display types are implemented with dictionaries that are managed by `Brain` and `_Hemisphere`. This also seems to lead to inconsistent API in which annotations are removed through `Brain.remove_annot()` while overlays are removed with `Brain.overlays_dict[x].remove()`. Is specialized classes for displays something we want to move towards or way from? While it would be easier to reduce reliance on the class (keep it for backwards compatibility or deprecate it) I think classes could make things somewhat more transparent, and possibly expose functions that make it easier for users to customize displays.

Eric89GXL:
>  all other display types are implemented with dictionaries that are managed by Brain and _Hemisphere

This sounds like it's more backend-agnostic, which is something that I'd like to promote. Think about swapping in e.g. a `matplotlib` rendering backend, it would be easier to do if things are managed by `Brain` and `_Hemisphere` right?

> annotations are removed through Brain.remove_annot() while overlays are removed with Brain.overlays_dict[x].remove()

Perhaps this is because people only ever thought about having a single `annot` at a time, whereas multiple overlays is a reasonable thing to use?

> Is specialized classes for displays something we want to move towards or way from?

Can you write a few lines of (1) the current API and (2) the proposed alternatives APIs that move toward/away from the specialized classes? I still can't quite see it.

mwaskom:
> While updating all the displays I noticed that overlays are handled in a way that is not quite consistent with other objects, by being implemented with two classes (one to represent data, one to display it) while all other display types are implemented with dictionaries that are managed by Brain and _Hemisphere. 

This is very, very old code. IIRC the motivation here was that overlays are a little weird in that you are mapping a single file to multiple mayavi piplines (as the positive and negative components are shown independently). That motivated creating a separate class for Overlays instead of just sticking the mayavi objects in a dictionary as is otherwise the case.

> Perhaps this is because people only ever thought about having a single annot at a time, whereas multiple overlays is a reasonable thing to use?

This is correct though less directly relevant to having a class define the display type — i.e. you can have multiple outputs from `add_data` but they are just stored in a list I think.

surfer/tests/test_viz.py
```diff
@@ -113,21 +113,22 @@ def test_brains():
 @requires_fsaverage
 def test_annot():
     """Test plotting of annot."""
-    mlab.options.backend = 'test'
+    mlab.options.backend = 'auto'
```
christianbrodbeck: Yes, it's a different problem – if I don't do it it looks like new arrays are not assigned to the data-source, I get:
```
$ nosetests test_viz:test_annot
2017-06-14 15:49:44.261 python[48695:54025567] modalSession has been exited prematurely - check for a reentrant call to endModalSession:
Exception occurred in traits notification handler for object: <mayavi.tools.filters.SetActiveAttributeFactory object at 0x13166a950>, trait: point_scalars, old value: , new value: 2
Traceback (most recent call last):
  File "/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/traits/trait_notifiers.py", line 340, in __call__
    self.handler( *args )
  File "/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/mayavi/tools/pipe_base.py", line 198, in _anytrait_changed
    setattr(obj, components[-1], value)
  File "/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/mayavi/core/trait_defs.py", line 97, in set_value
    raise TraitError(object, name,"one of %s"%values, value)
TraitError: The 'point_scalars_name' trait of a SetActiveAttribute instance must be one of [], but a value of '2' <type 'str'> was specified.
E
======================================================================
ERROR: Test plotting of annot.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/numpy/testing/decorators.py", line 147, in skipper_func
    return f(*args, **kwargs)
  File "/Users/christian/Code/PySurfer/surfer/tests/test_viz.py", line 122, in test_annot
    brain.add_annotation(a, b, p)
  File "/Users/christian/Code/PySurfer/surfer/viz.py", line 1229, in add_annotation
    brain['brain'].add_annotation(annot, ids, cmap))
  File "/Users/christian/Code/PySurfer/surfer/viz.py", line 2830, in add_annotation
    array_id, pipe = self._add_scalar_data(ids)
  File "/Users/christian/Code/PySurfer/surfer/viz.py", line 2727, in _add_scalar_data
    self._mesh_dataset, point_scalars=array_id, figure=self._f)
  File "/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/mayavi/tools/pipe_base.py", line 38, in the_function
    factory = factory_class(*args, **kwargs)
  File "/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/mayavi/tools/pipe_base.py", line 163, in __init__
    self.set(**traits)
  File "/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/mayavi/tools/pipe_base.py", line 180, in set
    self._anytrait_changed(trait, value)
  File "/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/mayavi/tools/pipe_base.py", line 198, in _anytrait_changed
    setattr(obj, components[-1], value)
  File "/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/mayavi/core/trait_defs.py", line 97, in set_value
    raise TraitError(object, name,"one of %s"%values, value)
TraitError: The 'point_vectors_name' trait of a SetActiveAttribute instance must be one of [], but a value of '' <type 'str'> was specified.
-------------------- >> begin captured stdout << ---------------------
Exception occurred in traits notification handler.
Please check the log file for details.

--------------------- >> end captured stdout << ----------------------
-------------------- >> begin captured logging << --------------------
apptools.preferences.preferences: DEBUG: loading preferences from </Users/christian/.enthought/mayavi_e3/preferences.ini>
apptools.preferences.preferences: DEBUG: loading preferences from <<open file '/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/mayavi/preferences/preferences.ini', mode 'rb' at 0x11da33ed0>>
apptools.preferences.preferences: DEBUG: loading preferences from <<open file '/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/tvtk/plugins/scene/preferences.ini', mode 'rb' at 0x11db008a0>>
mayavi.core.registry: DEBUG: Engine [<mayavi.core.null_engine.NullEngine object at 0x128e8ac50>] named NullEngine1 registered
mayavi.core.common: ERROR: No valid current object, please select an active object.
traits: ERROR: Exception occurred in traits notification handler for object: <mayavi.tools.filters.SetActiveAttributeFactory object at 0x13166a950>, trait: point_scalars, old value: , new value: 2
Traceback (most recent call last):
  File "/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/traits/trait_notifiers.py", line 340, in __call__
    self.handler( *args )
  File "/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/mayavi/tools/pipe_base.py", line 198, in _anytrait_changed
    setattr(obj, components[-1], value)
  File "/Users/christian/anaconda/envs/mayavi/lib/python2.7/site-packages/mayavi/core/trait_defs.py", line 97, in set_value
    raise TraitError(object, name,"one of %s"%values, value)
TraitError: The 'point_scalars_name' trait of a SetActiveAttribute instance must be one of [], but a value of '2' <type 'str'> was specified.
--------------------- >> end captured logging << ---------------------

----------------------------------------------------------------------
Ran 1 test in 2.321s

FAILED (errors=1)
```

Eric89GXL: Argh. Okay. If the CIs are okay with it I can live with it.

BTW I just pushed a fix to `master` that I made a few days ago to deal with the `imwrite` failure. I'll restart CIs on your latest commit so they hopefully pick it up.

christianbrodbeck:
After updating the overlay object I think this PR should not be affected the classes vs. dicts discussion, so if you want to keep it open I would move it into a separate issue; but here my thoughts:

> IIRC the motivation here was that overlays are a little weird in that you are mapping a single file to multiple mayavi piplines (as the positive and negative components are shown independently). That motivated creating a separate class for Overlays instead of just sticking the mayavi objects in a dictionary as is otherwise the case.

I guess these dictionaries grew over time in complexity... after adding a data-layer:

```
In [2]: brain.data_dict
Out[2]: 
{'lh': {'array': array([[ 1.43034017,  0.51767367,  0.45605102,  1.70507824],
         [ 0.26418859,  0.52732748,  0.90342617,  0.99659139],
         [ 1.78131807,  1.26297808,  1.32284141,  0.9324193 ],
         ..., 
         [ 2.65984082,  4.03173685,  6.1883111 ,  4.20004129],
         [ 2.9126246 ,  3.85630774,  3.45926285,  2.47084212],
         [ 3.17013526,  3.28790569,  2.09640741,  1.67996788]], dtype=float32),
  'colorbars': [<mayavi.core.lut_manager.LUTManager at 0x12b813950>],
  'fmax': 22.0,
  'fmid': 18.0,
  'fmin': 13.0,
  'layer': 0,
  'orig_ctable': array([[ 10,   0,   0, 255],
         [ 13,   0,   0, 255],
         [ 15,   0,   0, 255],
         ..., 
         [255, 255, 247, 255],
         [255, 255, 251, 255],
         [255, 255, 255, 255]], dtype=uint8),
  'smooth_mat': <163842x2562 sparse matrix of type '<type 'numpy.float64'>'
  	with 847182 stored elements in COOrdinate format>,
  'smoothing_steps': 10,
  'surfaces': [<mayavi.modules.surface.Surface at 0x138464d10>],
  'time': array([ 0.08,  0.09,  0.1 ,  0.11]),
  'time_idx': 0,
  'time_label': <function __main__.<lambda>>,
  'transparent': True,
  'vertices': array([   0,    2,    3, ..., 1348, 2451, 2452], dtype=uint32)},
 'rh': {'array': array([[ 1.1138016 ,  1.63564575,  1.56338835,  0.62299478],
         [ 4.47758293,  4.73766136,  3.3510077 ,  1.50541949],
         [ 0.35411292,  0.7056011 ,  0.65926534,  0.89739203],
         ..., 
         [ 4.18813515,  3.81059504,  1.91492558,  1.60494661],
         [ 3.19541049,  4.48996067,  3.88375044,  2.12374163],
         [ 3.23499441,  6.16087055,  6.44084167,  3.88419032]], dtype=float32),
  'colorbars': [<mayavi.core.lut_manager.LUTManager at 0x12ef84fb0>],
  'fmax': 22.0,
  'fmid': 18.0,
  'fmin': 13.0,
  'layer': 1,
  'orig_ctable': array([[ 10,   0,   0, 255],
         [ 13,   0,   0, 255],
         [ 15,   0,   0, 255],
         ..., 
         [255, 255, 247, 255],
         [255, 255, 251, 255],
         [255, 255, 255, 255]], dtype=uint8),
  'smooth_mat': <163842x2562 sparse matrix of type '<type 'numpy.float64'>'
  	with 847182 stored elements in COOrdinate format>,
  'smoothing_steps': 10,
  'surfaces': [<mayavi.modules.surface.Surface at 0x12ef84530>],
  'time': array([ 0.08,  0.09,  0.1 ,  0.11]),
  'time_idx': 0,
  'time_label': <function __main__.<lambda>>,
  'transparent': True,
  'vertices': array([   0,    2,    3, ..., 1348, 2451, 2452], dtype=uint32)}}
```

And this is only one layer... I assume dating back from when only one layer was possible. If there are multiple layers they are stored in `Brain._data_dicts` (which I made private when I added it to stop adding more attributes that have to be kept for backwards compatibility).

> Can you write a few lines of (1) the current API and (2) the proposed alternatives APIs that move toward/away from the specialized classes? I still can't quite see it.

What I was thinking about (analogous to overlays) was replacing those dictionaries with a `DataLayer` object that could have methods like `.scale_colormap()`. As such I think it would actually move towards backend-neutrality because it would interpose a layer of PySurfer API in front of the mayavi objects' API rather than exposing the mayavi objects themselves.

Eric89GXL:
> As such I think it would actually move towards backend-neutrality because it would interpose a layer of PySurfer API in front of the mayavi objects' API rather than exposing the mayavi objects themselves.

That sounds nice. But I agree it should be another PR if possible.

mwaskom:
> That sounds nice. But I agree it should be another PR if possible.

Agreed — it's a good idea but will be disruptive and will require a lot of thought to get right (which clearly wasn't done the first time, hence the mess).

surfer/utils.py
```diff
@@ -76,20 +104,32 @@ def __init__(self, subject_id, hemi, surf, subjects_dir=None,
...
+        nn = _compute_normals(coords, faces)
+
+        if self.coords is None:
+            self.coords = coords
+            self.faces = faces
+            self.nn = nn
+        else:
+            self.coords[:] = coords
+            self.faces[:] = faces
+            self.nn[:] = nn
```
christianbrodbeck: Since mesh objects now use the data stored on the `Surface` objects I changed this to modify it in place, assuming that `Surface` objects are not used independently; if you think that this is too much change I can make it optional through a parameter.

christianbrodbeck:
Tests pass.

Regarding naming – I preferred `.set_surface()` because it's not an abbreviation, and in `__init__` `surf` is a positional argument so the name is not highly salient... But if you prefer `set_surf()` I will update the name.

mwaskom:
> Regarding naming – I preferred .set_surface() because it's not an abbreviation, and in __init__ surf is a positional argument so the name is not highly salient...

In my experience, the term `surf` is very commonly used in Freesurfer-land as a parameter that takes the an argument of `white`, `pial`, `inflated`, etc. that identifies different representations of the triangular mesh. Whereas "surface" is a more general term for the 2D space that is differentiated from the "volume". But others can weigh in.

christianbrodbeck:
That makes sense, coming from the MNE side I did not know about this FreeSurfer convention...

christianbrodbeck:
Search-and-replace also found some `surface` in the examples, I replaced those too for consistency.

christianbrodbeck:
This is ready for review now. To test the new function use `Brain.set_surf(...)`.

surfer/viz.py
```diff
@@ -1745,6 +1751,39 @@ def set_distance(self, distance=None):
...
+
+        # update mesh objects (they use a reference to geo.coords)
+        with mayavi_threshold_patch:
+            for brain in self.brains:
+                brain.update_surf()
+
+        self.surf = surf
+        self._toggle_render(True, views)
+        for brain in self.brains:
+            brain._f.scene.reset_zoom()
```
christianbrodbeck: I wonder if this should be optional – but the geometries have slightly different centers, so without it, the brain seems to jump to the side.

Eric89GXL:
I wonder if we should (via deprecation) explicitly set the focal point to
0, 0, 0 instead. It would change interaction a little bit but perhaps in a
useful way.

christianbrodbeck:
And align the geometry so that 0, 0, 0 is the actual center?

Eric89GXL:
Ideally we wouldn't need to do any realignment assuming the FreeSurfer surfaces are defined properly with 0, 0, 0 to have some meaning. But maybe they aren't, in which case aligning to the origin maybe isn't such a good idea.

christianbrodbeck:
I think they aren't, there is a comment in the code under `Brain.set_view()` added by you :)

```
            # DO NOT set focal point, can screw up non-centered brains
            # view['focalpoint'] = (0.0, 0.0, 0.0)
```

Eric89GXL:
Whoops :) 

Let's leave the resetting in for now, and maybe we can think of a better solution later

surfer/viz.py
```diff
@@ -2682,19 +2717,67 @@ def _min_diff(self, beg, end):
...
+        self._mesh_dataset.point_data.get_array(array_id).name = array_id
+        self._mesh_dataset.point_data.update()
+
+        # build visualization pipeline
+        with warnings.catch_warnings(record=True):
+            pipe = mlab.pipeline.set_active_attribute(
+                self._mesh_dataset, point_scalars=array_id, figure=self._f)
+            # The new data-source is added to the wrong figure by default
+            # (a Mayavi bug??)
+            self._f.add_child(pipe.parent)
```
Eric89GXL: Can this be fool-proofed by something like `if not pipe.parent in self._f.children` or whatever the correct check would be?

mwaskom:
Good thinking, past Eric.

surfer/viz.py
```diff
@@ -2682,19 +2717,67 @@ def _min_diff(self, beg, end):
...
-        surf = OverlayDisplay(old, figure=self._f)
-        for bar in ["pos_bar", "neg_bar"]:
-            try:
-                self._format_cbar_text(getattr(surf, bar))
-            except AttributeError:
-                pass
-        return surf
+        array_id, mesh = self._add_scalar_data(old.mlab_data)
+
+        with mayavi_threshold_patch:
```
Eric89GXL: do all (or most) of these lines really need to be patched, or only a couple?

christianbrodbeck: The two `threshold` calls need to be patched, I moved the context up because that way it has to be entered only once. Do you prefer to patch the two calls separately?

surfer/utils.py
```diff
@@ -23,6 +26,31 @@
     string_types = str
 
 
+class _MayaviThresholdPatch(object):
+    """Context to monkey-patch Mayavi 4.5
+
+    In Mayavi 4.5, filters seem to be missing a .point_data attribute that
+    Threshold accesses on initialization.
+    """
+    need_patch = LooseVersion(mayavi.__version__) <= LooseVersion('4.5.0')
```
mwaskom: I'm a little confused with this is `<=`, shouldn't it be `==` based on the description above, which suggests a problem just on 4.5.0? That's the current version of mayavi so this will always be True, unless I'm missing something.

surfer/utils.py
```diff
@@ -32,7 +32,7 @@ class _MayaviThresholdPatch(object):
     In Mayavi 4.5, filters seem to be missing a .point_data attribute that
     Threshold accesses on initialization.
     """
-    need_patch = LooseVersion(mayavi.__version__) <= LooseVersion('4.5.0')
+    need_patch = LooseVersion(mayavi.__version__) == LooseVersion('4.5.0')
```
mwaskom: I should have been more clear in my previous review that I am ignorant about the issue here. When I suggested using `==` I was just saying that was implied by the docstring. Do we know the underlying issue will be fixed in Mayavi 4.5.1?

christianbrodbeck: It seemed to be fixed on master, so I'm assuming it will be fixed > 4.5. I don't know about mayavi 4.4, I tried reinstalling 4.4 with conda to try it out but hit upon error after error suggesting backend compatibility issues and gave up... :( does anybody still have a working 4.4 env?

CHANGES
```diff
@@ -1,6 +1,12 @@
 PySurfer Changes
 ================
 
+Version 0.8
+-----------
+
+- The surface geometry that is displayed can now be changed after initializing
+  a ``Brain`` instance with e.g. ``Brain.set_surf("smoothwm")``.
```
mwaskom: A tiny thing but it should probably by `brain.set_surf("smoothwm")` as it's an instance method.

mwaskom:
Uh oh; why did test coverage go down 24%?

christianbrodbeck:
That seems wrong – I pushed too early and then force pushed a few times to amend, maybe that confused coverage – Travis shows 72% for this PR as well as for master.

christianbrodbeck:
I successfully ran the tests with this env: `$ conda create -n mayavi44 numpy scipy matplotlib mayavi=4.4 h5py pyqt=4.10 pygments=1` so it looks like monkey-patching for `== 4.5` is right. I hope the new push will fix the coverage.

Eric89GXL:
This causes a freeze in MNE tests with "No valid current object" popup, can you try `nosetests mne/viz/tests/test_3d.py` and see if you can replicate?

Eric89GXL:
https://travis-ci.org/mne-tools/mne-python/jobs/246229590#L2360

Eric89GXL:
```
mayavi.core.common: ERROR: No valid current object, please select an active object.
traits: ERROR: Exception occurred in traits notification handler for object: <mayavi.tools.filters.SetActiveAttributeFactory object at 0x12f852ad0>, trait: point_scalars, old value: , new value: 2
Traceback (most recent call last):
  File "/Users/larsoner/anaconda/lib/python2.7/site-packages/traits/trait_notifiers.py", line 340, in __call__
    self.handler( *args )
  File "/Users/larsoner/anaconda/lib/python2.7/site-packages/mayavi/tools/pipe_base.py", line 198, in _anytrait_changed
    setattr(obj, components[-1], value)
  File "/Users/larsoner/anaconda/lib/python2.7/site-packages/mayavi/core/trait_defs.py", line 97, in set_value
    raise TraitError(object, name,"one of %s"%values, value)
TraitError: The 'point_scalars_name' trait of a SetActiveAttribute instance must be one of [], but a value of '2' <type 'str'> was specified.
```

christianbrodbeck:
If I recall correctly that was the error I got with the `test` backend, the reason for changing them to `auto`. It means the array the new data is not assigned to the data-source object (`2` is the name of the array). Can you try setting the backend to `auto`?

Eric89GXL:
> Can you try setting the backend to auto?

Even if it fixes it, I don't think it's a good solution to the problem. This implies that these changes propagate problems to other packages, so it would be better if we could find some fix that doesn't do this.

christianbrodbeck:
This skips the parts that are broken under the testing backend, not great but it should work for now – the mne-python tests pass... I hope https://github.com/enthought/mayavi/pull/514 turns out to be a valid fix

Eric89GXL:
@mwaskom do you have time to look?

mwaskom:
Thanks @christianbrodbeck!

PR #195, opened 2017-06-25
ENH: Add doc build

Eric89GXL:
This should fix AppVeyor and make it build the docs (in noplot mode).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: c
-dist: trusty
 sudo: false
 # This list of requirements is way overkill, but it helps ensure we
 # actually load a usable config (it works for MNE-Python)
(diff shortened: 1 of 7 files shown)


Eric89GXL:
I'll merge this once the CIs are happy since it fixes AppVeyor

christianbrodbeck:
Is reverting the mlab backend necessary to get the tests to work? It will make the tests less complete...

Eric89GXL:
They cause AppVeyor to crash

christianbrodbeck:
I see... could we make it system-dependent?

Eric89GXL:
... they also cause windows to open locally during testing, which is less than ideal. The best thing would be to make Mayavi more fully support the `test` backend.

But I agree it's slightly better to cover more fully, so I'll make a commit to triage for Windows.

christianbrodbeck:
Yeah I agree windows opening during testing are less than ideal...  at least as it is now you can change to the `test` backend temporarily with one change.

Eric89GXL:
Okay we're green so I'll squash and merge. But if people feel like looking feel free and we can iterate.

PR #196, opened 2017-06-27
FIX:  negative overlay

christianbrodbeck:
Bug introduced in https://github.com/nipy/PySurfer/pull/191


Eric89GXL:
Thanks @christianbrodbeck

PR #197, opened 2017-06-29
FIX Brain.set_data_time_index():  skip data without time axis

christianbrodbeck:
Fix for a corner case where multiple data layers are added, and not all of them have a time axis

### DIFF
--- a/surfer/tests/test_viz.py
+++ b/surfer/tests/test_viz.py
@@ -264,6 +264,11 @@ def test_meg_inverse():
     assert_equal(data_dicts[0]['time_idx'], 0)
     assert_equal(data_dicts[1]['time_idx'], 0)
 
+    # add second data-layer without time axis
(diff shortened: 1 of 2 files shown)


mwaskom:
Ah I think I've hit this before and couldn't figure out why. Thanks!

mwaskom:
Does anyone know why our travis builds are taking forever to get going lately?

Eric89GXL:
No idea. It has been slower to run but not to start on other repos

mwaskom:
Is this travis failure related to these changes?

christianbrodbeck:
Very unlikely it's form these changes because the relevant test (`test_image `) never adds a data-layer. Also could not spot any changed version in the installed packages. I can't rerun the build to see whether it fails consistently, can one of you?

mwaskom:
Kicking it to see what happens...

christianbrodbeck:
all clear...

Issue #200, opened 2017-07-12
Medial view non-"region" is being mapped onto


YSanchezAraujo:
In the picture below, I think the large area in the middle of the medial views is usually greyed out in images and in this use case (I think it should just be white)I'm not sure why values are being mapped onto it. 

![xb_brain](https://user-images.githubusercontent.com/17433891/28125636-29a050a0-66f5-11e7-859e-004fdd718707.png)

That result is gotten using: 

```python
import os
import sys
import numpy as np
from custom import utils
import nibabel as nb
from surfer import Brain

result_name = sys.argv[1]
header_name = sys.argv[2]
save_name = sys.argv[3]

brain = Brain("fsaverage", "split", "inflated",
               views=['lat', 'med'], background="white")


hp = "/storage/gablab001/data/genus/fs_cog/exp1/column_headers"
rp = "/storage/gablab001/data/genus/fs_cog/exp1/lg/results"
with open(os.path.join(hp, header_name), "r") as tmp:
    headers = np.array(tmp.read().strip("\n").split("\n"))

res = utils.read_pickle(os.path.join(rp,"{}".format(result_name)))
coefs = [val for key, val in res.items() if "coef" in key]
nz = lambda x: np.nonzero(x)[0]
features_selected_lh = []
features_selected_rh = []
features_all = []

from collections import Counter
import pandas as pd

for coef in coefs:
    idx = nz(coef)
    features_all.extend(headers[idx].tolist())
    for col in headers[idx].tolist():
        if 'lh' in col:
            features_selected_lh.append(col)
        elif 'rh' in col:
            features_selected_rh.append(col)

left_count = dict(Counter(features_selected_lh))
right_count = dict(Counter(features_selected_rh))

left_df = pd.DataFrame({
    'col': [key.replace('lh_','').replace('.','-').replace('_thickness_D','')
            for key, val in left_count.items()],
    'val': [val for key, val in left_count.items()]
})

right_df = pd.DataFrame({
    'col': [key.replace('rh_','').replace('.','-').replace('_thickness_D','')
            for key, val in right_count.items()],
    'val': [val for key, val in right_count.items()]
})

label_dir = '/cm/shared/openmind/freesurfer/5.3.0/subjects/fsaverage/label'
left_label_file = 'lh.aparc.a2009s.annot'
right_label_file = 'rh.aparc.a2009s.annot'
lh_aparc_file = os.path.join(label_dir, left_label_file)
[...]

PR #201, opened 2017-07-12
Handle undefind vertices in parcellation example

mwaskom:
#200 reminded me that by switching the parcellation example back to using the aparc, it's necessary to handle undefined vertices.


Eric89GXL:
@mwaskom good to go from your end?

mwaskom:
Yep, simple change, results look fine.

PR #203, opened 2017-07-15
Colormap as list tweaks

christianbrodbeck:
- Allow color-lists with colors including alpha
- Allow specifying color-lists as tuples (which was possible previously, so since tuples are often used interchangeably with lists I am not sure whether you removed this intentionally)

### DIFF
--- a/surfer/utils.py
+++ b/surfer/utils.py
@@ -497,7 +497,7 @@ def create_color_lut(cmap, n_colors=256):
     surfer_cmaps += [name + "_r" for name in surfer_cmaps]
 
     if not isinstance(cmap, string_types) and isinstance(cmap, Sequence):
-        colors = list(map(mpl.colors.colorConverter.to_rgb, cmap))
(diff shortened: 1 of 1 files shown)


Eric89GXL:
LGTM, @mwaskom do you want to look since this touches some code you modified last week?

mwaskom:
Seems fine; I assume that the behavior of checking against `collections.Sequence` is invariant across Python 2/3?

Given the already confusing difference between in how we handle arrays/lists it felt a little cleaner to be strict that the list behavior applies explicitly to lists and not "list-like things", but I guess I don't have a problem accepting tuples.

christianbrodbeck:
I'm not aware of any relevant changes ([Python docs](https://docs.python.org/3/library/collections.abc.html?highlight=sequence#collections.abc.Sequence)). If you want to be more conservative we could use `(list, tuple)` in the type check, but I think `Sequence` would be appropriate for Python duck typing.

PR #204, opened 2017-07-20
MRG: Add support for vector data in `add_data`

Eric89GXL:
Eventually this will let us add vector data like `(n_vertices, 3, n_time)`, which is useful for electrophysiological estimates (e.g., MEG inverse solutions). Right now it doesn't look so good but it at least does not throw an error.

Code adapted from the [MNE PR](https://github.com/mne-tools/mne-python/pull/3842) by @wmvanvliet.

Todo:

- [x] do not upsample the vectors to the high-res mesh
- [x] check `set_time_index`
- [x] fix `scale_data_colormap`
- [x] ~~split from `add_data` to `add_vectors`~~
- [x] ~~add optional `nn=None` argument~~
- [x] fix TimeViewer / `set_data_smoothing_steps`
- [x] check data addition/removal
- [x] check `time_viewer`
- [x] deal with any outstanding XXX's in code
- [x] add/test brain surface transparency (glass brain)
- [x] make sure everything is pretty
- [x] add tests

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -996,7 +996,7 @@ class Brain(object):
             or a custom look up table (an n x 4 array coded with RBGA values
             between 0 and 255).
         alpha : float in [0, 1]
-            alpha level to control opacity of the surface.
+            alpha level to control opacity of the overlay.
         vertices : numpy array
             vertices for which the data is defined (needed if len(data) < nvtx)
         smoothing_steps : int or None
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -965,28 +965,26 @@ class Brain(object):
                  vertices=None, smoothing_steps=20, time=None,
                  time_label="time index=%d", colorbar=True,
                  hemi=None, remove_existing=False, time_label_size=14,
-                 initial_time=None):
+                 initial_time=None, scale_factor=None):
         """Display data from a numpy array on the surface.
 
-        This provides a similar interface to add_overlay, but it displays
+        This provides a similar interface to
+        :meth:`surfer.viz.Brain.add_overlay`, but it displays
         it with a single colormap. It offers more flexibility over the
-        colormap, and provides a way to display four dimensional data
-        (i.e. a timecourse).
+        colormap, and provides a way to display four-dimensional data
+        (i.e., a timecourse) or five-dimensional data (i.e., a
+        vector-valued timecourse).
 
-        Note that min sets the low end of the colormap, and is separate
-        from thresh (this is a different convention from add_overlay)
-
-        Note: If the data is defined for a subset of vertices (specified
-        by the "vertices" parameter), a smoothing method is used to interpolate
-        the data onto the high resolution surface. If the data is defined for
-        subsampled version of the surface, smoothing_steps can be set to None,
-        in which case only as many smoothing steps are applied until the whole
-        surface is filled with non-zeros.
+        .. note:: ``min`` sets the low end of the colormap, and is separate
+                  from thresh (this is a different convention from
+                  :meth:`surfer.viz.Brain.add_overlay`).
 
         Parameters
         ----------
-        array : numpy array
-            data array (nvtx vector)
+        array : numpy array, shape (n_vertices[, 3][, n_times])
+            Data array. For the data to be understood as vector-valued
+            (3 values per vertex), then ``array`` must be have all 3
+            dimensions.
         min : float
             min value in colormap (uses real min if None)
         max : float
@@ -998,7 +996,7 @@ class Brain(object):
             or a custom look up table (an n x 4 array coded with RBGA values
             between 0 and 255).
         alpha : float in [0, 1]
-            alpha level to control opacity
+            alpha level to control opacity of the surface.
         vertices : numpy array
             vertices for which the data is defined (needed if len(data) < nvtx)
         smoothing_steps : int or None
@@ -1023,6 +1021,18 @@ class Brain(object):
         initial_time : float | None
             Time initially shown in the plot. ``None`` to use the first time
             sample (default).
+        scale_factor : float | None (default)
+            The scale factor to use when displaying glyphs for vector-valued
+            data.
+
+        Notes
+        -----
+        If the data is defined for a subset of vertices (specified
+        by the "vertices" parameter), a smoothing method is used to interpolate
+        the data onto the high resolution surface. If the data is defined for
+        subsampled version of the surface, smoothing_steps can be set to None,
+        in which case only as many smoothing steps are applied until the whole
+        surface is filled with non-zeros.
         """
         hemi = self._check_hemi(hemi)
         array = np.asarray(array)
@@ -1031,6 +1041,7 @@ class Brain(object):
             min = array.min() if array.size > 0 else 0
         if max is None:
             max = array.max() if array.size > 0 else 0
+        mid = (min + max) / 2.
 
         # Create smoothing matrix if necessary
         if len(array) < self.geo[hemi].x.shape[0]:
@@ -1042,19 +1053,25 @@ class Brain(object):
         else:
             smooth_mat = None
 
-        # Calculate initial data to plot
-        if array.ndim == 1:
-            array_plot = array
-        elif array.ndim == 2:
-            array_plot = array[:, 0]
-        else:
-            raise ValueError("data has to be 1D or 2D")
-
-        if smooth_mat is not None:
-            array_plot = smooth_mat * array_plot
-
-        # Copy and byteswap to deal with Mayavi bug
-        mlab_plot = _prepare_data(array_plot)
+        magnitude = None
+        magnitude_max = None
+        if array.ndim == 3:
+            if array.shape[1] != 3:
+                raise ValueError('If array has 3 dimensions, array.shape[1] '
+                                 'must equal 3, got %s' % (array.shape[1],))
+            magnitude = np.linalg.norm(array, axis=1)
+            if scale_factor is None:
+                distance = np.sum([array[:, dim, :].ptp(axis=0).max() ** 2
+                                   for dim in range(3)])
+                if distance == 0:
+                    scale_factor = 1
+                else:
+                    scale_factor = (0.4 * distance /
+                                    (4 * array.shape[0] ** (0.33)))
+            magnitude_max = magnitude.max()
+        elif array.ndim not in (1, 2):
+            raise ValueError('array has must have 1, 2, or 3 dimensions, '
+                             'got (%s)' % (array.ndim,))
 
         # Process colormap argument into a lut
         lut = create_color_lut(colormap)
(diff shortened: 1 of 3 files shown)


wmvanvliet:
Awesome!

Eric89GXL:
Okay this is ready for review/merge from my end. AppVeyor is unhappy because I'm in the midst of reconfiguring it. Here is the example output, which is not a pretty as what @wmvanvliet has posted previously because this isn't a free-orientation solution (but we can add it later):

snapshot

I also snuck in fixes for normal-setting (needed `.update()` calls) and `set_data_smoothing_steps`, which are both broken in `master`. I also enabled backface culling for the brain surface, which seems to clean up the translucent plot quite a bit. But I can roll that back (or make it optional) if it's not agreed to be better.

wmvanvliet:
I'll test it first this weekend when I get a chance

wmvanvliet:
Hrm, not really working for me. This is the output I get:

screen shot 2017-07-23 at 20 54 46

From this script: 
https://gist.github.com/wmvanvliet/a96a9b883cf492bed807559d51abe310

wmvanvliet:
[cancel my last comment, it's working fine when I scale the colormap properly]

Eric89GXL:
So +1 for merge or do you need to look or test more?

mwaskom:
I don't know anything about the kind of data you're working with here, but the new translucent cortex does look very nice!

wmvanvliet:
I want to test a little more.

wmvanvliet:
There seems to be a problem with the transparent arrow glyphs. They are transparent to the cortex surface, but not to the data plotted on the surface... They may be a rendering order thing. I remember encountering it a while back and having to play with the ordering.

Eric89GXL:
Right now I don't set the `alpha` of the arrows to anything (so they are always 1.). Should I set their alpha to match that of the data surface instead? Maybe that is what you see?

Eric89GXL:
Commit pushed to split vector alpha.

@wmvanvliet does it fix your issue?

wmvanvliet:
Setting `vector_alpha` < 1.0 does indeed solve the transparency issue, but I'd like my vectors to be solid, and setting `vector_alpha=1.0` still shows the same bug.

wmvanvliet:
As a hack setting `vector_alpha=0.999999` seems to work fine. Strange.

wmvanvliet:
`brain.set_time()` breaks the transparency of the vectors. This may be related to the bug.

Eric89GXL:
@wmvanvliet can you detail how to see this breakage?

Eric89GXL:
(and post a screenshot with the problem)

wmvanvliet:
Running plot_vector_mne_solution.py from https://github.com/mne-tools/mne-python/pull/3842 shows gray arrows that should be transparent:
screenshot1

Calling `brain.set_time()` resets the scaling of the colormap of the arrows:
screenshot2

Eric89GXL:
Okay @wmvanvliet both problem solved, the first by forcing 1->0.99999, and the second by setting the `lut` and `data_range` after any calls to `update` (as it resets them). Can you see if it works for you now?

wmvanvliet:
Beautiful! Everything works on my end now.

screenshot

wmvanvliet:
movie

agramfort:
https://twitter.com/agramfort/status/890665952362110976 :)

Eric89GXL:
@mwaskom looks like the MEG folks are happy with the functionality, do you have time to look at the code to make sure it fits PySurfer properly?

surfer/viz.py
```diff
@@ -998,7 +996,7 @@ def add_data(self, array, min=None, max=None, thresh=None,
             or a custom look up table (an n x 4 array coded with RBGA values
             between 0 and 255).
         alpha : float in [0, 1]
-            alpha level to control opacity
+            alpha level to control opacity of the surface.
```
mwaskom: This is a little confusing since it controls the opacity of the statistical overlay w/r/t whatever is plotted under it, not the mesh itself (which I assume most would think of as the "surface") right?

surfer/viz.py
```diff
@@ -965,28 +965,26 @@ def add_data(self, array, min=None, max=None, thresh=None,
...
-        Note that min sets the low end of the colormap, and is separate
-        from thresh (this is a different convention from add_overlay)
-
-        Note: If the data is defined for a subset of vertices (specified
-        by the "vertices" parameter), a smoothing method is used to interpolate
-        the data onto the high resolution surface. If the data is defined for
-        subsampled version of the surface, smoothing_steps can be set to None,
-        in which case only as many smoothing steps are applied until the whole
-        surface is filled with non-zeros.
+        .. note:: ``min`` sets the low end of the colormap, and is separate
```
mwaskom: Why the two different approaches to reformatting the notes?

mwaskom:
Personally it feels like a lot of extra complexity is being stuffed into the poor `add_data` method, but I can live with it if it's important to the MEG folks.

Going forward it might be worth thinking about a path towards more focused interfaces to plotting 4D/5D data.

surfer/viz.py
Eric89GXL: This one seemed to be more likely to cause problems / be higher priority than the other one that was moved to `Notes`

surfer/viz.py
Eric89GXL: I'll change to "overlay"

Eric89GXL:
> Going forward it might be worth thinking about a path towards more focused interfaces to plotting 4D/5D data.

Yeah I agree it adds some complexity. I had initially tried adding a new `add_vectors` method, but it ended up introducing some redundancy. Do you think it would be cleaner that way, though?

mwaskom:
I only gave it a small amount of thought but did see how it would be hard to avoid redundancy without completely reworking things, so I'm not overly bothered by this as a current solution.

Eric89GXL:
> I'm not overly bothered by this as a current solution.

Nice (unintentional) pun in there, since we're talking about visualizing a minimum-norm solution / estimates for neural currents...

Any objection to merge? If not let's get it in this weekend so @wmvanvliet can move forward with the MNE improvements.

mwaskom:
> Any objection to merge? 

No resistance on my end.

agramfort:
>
> No resistance on my end.
>
>
same for me

Eric89GXL:
@christianbrodbeck do you want to have a look? You've worked with `mayavi` internals a lot, and might be interested in the vector viz, too.

surfer/viz.py
```diff
@@ -965,28 +965,26 @@ def add_data(self, array, min=None, max=None, thresh=None,
...
 
-        This provides a similar interface to add_overlay, but it displays
+        This provides a similar interface to
+        :meth:`surfer.viz.Brain.add_overlay`, but it displays
         it with a single colormap. It offers more flexibility over the
-        colormap, and provides a way to display four dimensional data
-        (i.e. a timecourse).
+        colormap, and provides a way to display four-dimensional data
+        (i.e., a timecourse) or five-dimensional data (i.e., a
+        vector-valued timecourse).
```
christianbrodbeck: Isn't the vector data strictly speaking 7 dimensional, or 6 if you discount magnitude? :)

surfer/viz.py
```diff
@@ -965,28 +965,26 @@ def add_data(self, array, min=None, max=None, thresh=None,
...
+                  :meth:`surfer.viz.Brain.add_overlay`).
 
         Parameters
         ----------
-        array : numpy array
-            data array (nvtx vector)
+        array : numpy array, shape (n_vertices[, 3][, n_times])
+            Data array. For the data to be understood as vector-valued
+            (3 values per vertex), then ``array`` must be have all 3
+            dimensions.
```
christianbrodbeck: It might be worth adding the coordinate system for vectors (e.g. RAS)?

surfer/viz.py
```diff
@@ -1905,19 +1903,34 @@ def set_data_time_index(self, time_idx, interpolation='quadratic'):
...
 
+                vector_values = scalar_data.copy()
+                if data['smooth_mat'] is not None:
+                    scalar_data = data['smooth_mat'] * scalar_data
                 for brain in self.brains:
                     if brain.hemi == hemi:
-                        brain.set_data(data['layer_id'], plot_data)
+                        brain.set_data(data['layer_id'], scalar_data,
+                                       vectors, vector_values)
+                del brain
```
christianbrodbeck: is this `del` necessary?

surfer/viz.py
```diff
@@ -2813,18 +2936,40 @@ def add_overlay(self, old):
...
+        scale_factor_norm = None
+        if array.ndim == 3:
+            scale_factor_norm = scale_factor / magnitude_max
+            vectors = array[:, :, 0].copy()
+            glyphs = self._add_vector_data(
+                vectors, vector_values, fmin, fmid, fmax,
+                scale_factor_norm, vertices, vector_alpha, lut)
+        else:
+            glyphs = None
+        del scale_factor
```
christianbrodbeck: why `del`?

surfer/viz.py
```diff
@@ -1115,27 +1140,30 @@ def time_label(x):
...
-                    array, mlab_plot, min, max, thresh, lut, colormap, alpha,
-                    colorbar, layer_id)
+                s, ct, bar, gl = brain['brain'].add_data(
+                    array, min, mid, max, thresh, lut, colormap, alpha,
+                    colorbar, layer_id, smooth_mat, magnitude, magnitude_max,
+                    scale_factor, vertices, vector_alpha)
                 surfs.append(s)
                 bars.append(bar)
-                if array.ndim == 2 and time_label is not None:
+                glyphs.append(gl)
```
christianbrodbeck: Wouldn't it be simpler to just set `glyphs` to `None` or `()` instead of creating list of `None` for non-vector data?

Eric89GXL:
> Plotting vector data without time axis is currently somewhat complicated (e.g. setting time_label=None), might it be worth to illustrate it in the method docstring?

I think it might be YAGNI, at least to start. Once people without time axes want to start using it we can make it clearer :)

surfer/viz.py
Eric89GXL: it's mostly there for sanity check / stupid-coding prevention

surfer/viz.py
Eric89GXL: Same as above, here IIRC it was to prevent using something later that shouldn't be used

christianbrodbeck:
I see

surfer/viz.py
Eric89GXL: Heh, depends on what we mean by "dimensions" I guess

Eric89GXL:
I added the note anyway since it didn't take too much, @christianbrodbeck can you check to see if it's sufficient?

christianbrodbeck:
Yes looks good!

Eric89GXL:
Okay, I'll merge once CIs are happy, thanks for the reviews -- and @wmvanvliet for doing most of the hard work with getting it working in the first place

PR #205, opened 2017-07-31
MRG: Improve doc

### LINKED ISSUE #199: Issues with new sphinx-gallery examples
I just noticed to issues with the new example gallery:

- We have a few examples that don't produce an image (the ones where the script name does not start with `plot_*`, which was originally how we identified which scripts should populate the gallery)
- Examples that produce a multi-panel figure no longer show the multi-panel image that gets saved; now they show four separate images and only one becomes the thumbnail.

larsoner:
Closes #199.

Todo:
- [x] [alphabetical by filename](https://github.com/sphinx-gallery/sphinx-gallery/pull/281)
- [x] [250x250 thumbnails](https://github.com/sphinx-gallery/sphinx-gallery/issues/282)

~~Still need to make sure the rest of the issues in there are addressed.~~ In the meantime, the first commit fixes the vector example build (wouldn't work on OSX -- figures are closed at the end of each SG section).

### DIFF
--- a/surfer/tests/test_viz.py
+++ b/surfer/tests/test_viz.py
@@ -75,10 +75,13 @@ def test_image():
     brain = Brain(*std_args, size=100)
     brain.save_image(tmp_name)
     brain.save_image(tmp_name, 'rgba', True)
-    brain.save_montage(tmp_name, ['l', 'v', 'm'], orientation='v')
-    brain.save_montage(tmp_name, ['l', 'v', 'm'], orientation='h')
-    brain.save_montage(tmp_name, [['l', 'v'], ['m', 'f']])
     brain.screenshot()
+    if not sys.platform.startswith('linux') or \
+            os.getenv('TRAVIS', 'false') != 'true':
+        # for some reason these fail on Travis sometimes
+        brain.save_montage(tmp_name, ['l', 'v', 'm'], orientation='v')
+        brain.save_montage(tmp_name, ['l', 'v', 'm'], orientation='h')
+        brain.save_montage(tmp_name, [['l', 'v'], ['m', 'f']])
     brain.close()
 
 
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -969,7 +969,7 @@ class Brain(object):
         """Display data from a numpy array on the surface.
 
         This provides a similar interface to
-        :meth:`surfer.viz.Brain.add_overlay`, but it displays
+        :meth:`surfer.Brain.add_overlay`, but it displays
         it with a single colormap. It offers more flexibility over the
         colormap, and provides a way to display four-dimensional data
         (i.e., a timecourse) or five-dimensional data (i.e., a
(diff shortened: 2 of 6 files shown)


Eric89GXL:
Ready for review/merge from my end. Updated docs [here](https://ci.appveyor.com/api/buildjobs/r83cj2jfj0a77477/artifacts/doc%2F_build%2Fhtml.zip).

Eric89GXL:
Setting back to WIP because the SG names may change. But still ready for review / perusal of the doc because that shouldn't change.

larsoner:
Okay this one should be good to go. Includes a small fix for Travis.

surfer/tests/test_viz.py
```diff
@@ -75,10 +75,13 @@ def test_image():
     brain = Brain(*std_args, size=100)
     brain.save_image(tmp_name)
     brain.save_image(tmp_name, 'rgba', True)
-    brain.save_montage(tmp_name, ['l', 'v', 'm'], orientation='v')
-    brain.save_montage(tmp_name, ['l', 'v', 'm'], orientation='h')
-    brain.save_montage(tmp_name, [['l', 'v'], ['m', 'f']])
     brain.screenshot()
+    if not sys.platform.startswith('linux') or \
+            os.getenv('TRAVIS', 'false') != 'true':
```
agramfort: so this will not run on any linux system ?

larsoner: It runs fine on my local Linux, I'm assuming it's an Xvfb / temporary Mayavi bug

larsoner: if you're asking about the conditional -- these lines will run if the system is not linux or it's not travis (same as `not (linux and travis)`)

agramfort:
ok +1 for MRG then

larsoner:
Okay with you @mwaskom?

surfer/viz.py
```diff
@@ -1186,7 +1186,7 @@ def add_annotation(self, annot, borders=True, alpha=1, hemi=None,
             hemisphere, i.e. ``annot=(labels, ctab)`` for a single hemisphere
             or ``annot=((lh_labels, lh_ctab), (rh_labels, rh_ctab))`` for both
             hemispheres. ``labels`` and ``ctab`` should be arrays as returned
-            by :func:`nibabel.freesurfer.read_annot`.
+            by :func:`nibabel.freesurfer.io.read_annot`.
```
mwaskom: Is nibabel changing its imports? `nib.freesurfer.read_annot` has always worked for me.

mwaskom:
I don't have time to build the docs and look but the code changes seem good. Note the one comment.

surfer/viz.py
larsoner: It is probably imported both places, but this is where it lives in their docs (sphinx index) so this is necessary for it to link properly.

larsoner:
Thanks for the looks, I'll work on cutting a release now

surfer/viz.py
mwaskom: Ah got it.

PR #207, opened 2017-09-22
MRG+1: allowing transparency in divergent colormaps

sbitzer:
When analysing some of my data I came across the issue that I couldn't use a divergent colormap for, e.g., correlations or t-values and at the same time let small values in the middle of the colormap be transparent. So I have adapted Brain.scale_data_colormap such that it can now handle transparency in divergent colormaps properly. Also, I added the option to set the overall alpha-level of the colormap while maintaining transparency of a particular region in the colormap. The usage of these features is demonstrated in examples/plot_resting_correlations.py.

I believe that these changes make it easier to explore suitable visualisations of divergent data and hope that you find them useful!

### DIFF
--- a/examples/plot_resting_correlations.py
+++ b/examples/plot_resting_correlations.py
@@ -46,14 +46,18 @@ for overlay in brain.overlays_dict["ang_corr_rh"]:
 We want to use an appropriate color map for these data: a divergent map that
 is centered on 0, which is a meaningful transition-point as it marks the change
 from negative correlations to positive correlations.
-
-We'll also plot the map with some transparency so that we can see through to
-the underlying anatomy.
 """
-brain.add_data(surf_data_lh, -.7, .7, colormap="vlag", alpha=.75,
-               hemi='lh')
-brain.add_data(surf_data_rh, -.7, .7, colormap="vlag", alpha=.75,
-               hemi='rh')
+brain.add_data(surf_data_lh, -.7, .7, colormap="vlag", hemi='lh')
+brain.add_data(surf_data_rh, -.7, .7, colormap="vlag", hemi='rh')
+
+"""
+You can make the small correlations transparent by re-scaling the colormap with
+transparency and letting the rescaler know that this is a divergent color map.
+In this case the first number gives the value that should be in the center of
+the colormap, the second determines which values will be transparent and the 
+third determines the left and right edges of the new colormap.
+"""
+brain.scale_data_colormap(0, .35, .7, transparent=True, divergent=True)
 
 """
 This overlay represents resting-state correlations with a
--- a/examples/plot_resting_correlations.py
+++ b/examples/plot_resting_correlations.py
@@ -60,6 +60,13 @@ third determines the left and right edges of the new colormap.
 brain.scale_data_colormap(0, .35, .7, transparent=True, divergent=True)
 
 """
+You can also set the overall opacity of the displayed data while maintaining
+the transparency of the small values.
+"""
+brain.scale_data_colormap(0, .35, .7, transparent=True, divergent=True, 
+                          alpha=0.75)
+
+"""
 This overlay represents resting-state correlations with a
 seed in left angular gyrus. Let's plot that seed.
 """
--- a/examples/plot_resting_correlations.py
+++ b/examples/plot_resting_correlations.py
@@ -55,7 +55,7 @@ brain.add_data(surf_data_rh, 0, .7, center=0, hemi='rh')
 You can tune the data display by shifting the colormap around interesting
 regions. For example, you can ignore small correlation up to a magnitude of 0.2
 and let colors become gradually less transparent from 0.2 to 0.5 by re-scaling
-the colormap as follows. For more information see the help string of this 
+the colormap as follows. For more information see the help string of this
 function.
 """
 brain.scale_data_colormap(.2, .5, .7, transparent=True, center=0)
@@ -64,7 +64,7 @@ brain.scale_data_colormap(.2, .5, .7, transparent=True, center=0)
 You can also set the overall opacity of the displayed data while maintaining
 the transparency of the small values.
 """
-brain.scale_data_colormap(0, .35, .7, transparent=True, center=0, 
+brain.scale_data_colormap(0, .35, .7, transparent=True, center=0,
                           alpha=0.75)
 
 """
--- a/examples/plot_resting_correlations.py
+++ b/examples/plot_resting_correlations.py
@@ -45,25 +45,26 @@ for overlay in brain.overlays_dict["ang_corr_rh"]:
 """
 We want to use an appropriate color map for these data: a divergent map that
 is centered on 0, which is a meaningful transition-point as it marks the change
-from negative correlations to positive correlations.
+from negative correlations to positive correlations. By providing the 'center'
+argument the add_data function automatically chooses a divergent colormap.
 """
-brain.add_data(surf_data_lh, -.7, .7, colormap="vlag", hemi='lh')
-brain.add_data(surf_data_rh, -.7, .7, colormap="vlag", hemi='rh')
+brain.add_data(surf_data_lh, 0, .7, center=0, hemi='lh')
+brain.add_data(surf_data_rh, 0, .7, center=0, hemi='rh')
 
 """
-You can make the small correlations transparent by re-scaling the colormap with
-transparency and letting the rescaler know that this is a divergent color map.
-In this case the first number gives the value that should be in the center of
-the colormap, the second determines which values will be transparent and the 
-third determines the left and right edges of the new colormap.
+You can tune the data display by shifting the colormap around interesting
+regions. For example, you can ignore small correlation up to a magnitude of 0.2
+and let colors become gradually less transparent from 0.2 to 0.5 by re-scaling
+the colormap as follows. For more information see the help string of this 
+function.
 """
-brain.scale_data_colormap(0, .35, .7, transparent=True, divergent=True)
+brain.scale_data_colormap(.2, .5, .7, transparent=True, center=0)
 
 """
 You can also set the overall opacity of the displayed data while maintaining
 the transparency of the small values.
 """
-brain.scale_data_colormap(0, .35, .7, transparent=True, divergent=True, 
+brain.scale_data_colormap(0, .35, .7, transparent=True, center=0, 
                           alpha=0.75)
 
 """
--- a/surfer/utils.py
+++ b/surfer/utils.py
@@ -482,6 +482,7 @@ def create_color_lut(cmap, n_colors=256, center=None):
     center : double, optional
         indicates whether desired colormap should be for divergent values, 
         currently only used to select default colormap for cmap='auto'
+        
     Returns
     -------
     lut : n_colors x 4 integer array
@@ -499,7 +500,7 @@ def create_color_lut(cmap, n_colors=256, center=None):
 
             return lut
 
-    if isinstance(cmap, string_types) and cmap=="auto":
+    if isinstance(cmap, string_types) and cmap == "auto":
         if center is None:
             cmap = "rocket"
         else:
--- a/surfer/utils.py
+++ b/surfer/utils.py
@@ -500,6 +500,8 @@ def create_color_lut(cmap, n_colors=256, center=None):
 
             return lut
 
+    # choose default colormaps (REMEMBER to change doc, e.g., in 
+    # Brain.add_data, when changing these defaults)
     if isinstance(cmap, string_types) and cmap == "auto":
         if center is None:
             cmap = "rocket"
--- a/surfer/utils.py
+++ b/surfer/utils.py
@@ -464,7 +464,7 @@ def mesh_edges(faces):
     return edges
 
 
-def create_color_lut(cmap, n_colors=256):
+def create_color_lut(cmap, n_colors=256, center=None):
     """Return a colormap suitable for setting as a Mayavi LUT.
 
     Parameters
@@ -473,9 +473,15 @@ def create_color_lut(cmap, n_colors=256):
         Input colormap definition. This can be the name of a matplotlib
         colormap, a list of valid matplotlib colors, or a suitable
         mayavi LUT (possibly missing the alpha channel).
+        
+        if value is "auto", a default sequential or divergent colormap is 
+        returned
     n_colors : int, optional
         Number of colors in the resulting LUT. This is ignored if cmap
         is a 2d array.
+    center : double, optional
+        indicates whether desired colormap should be for divergent values, 
+        currently only used to select default colormap for cmap='auto'
     Returns
     -------
     lut : n_colors x 4 integer array
@@ -492,6 +498,12 @@ def create_color_lut(cmap, n_colors=256):
                 lut = np.c_[cmap, alpha]
 
             return lut
+
+    if isinstance(cmap, string_types) and cmap=="auto":
+        if center is None:
+            cmap = "rocket"
+        else:
+            cmap = "icefire"
 
     surfer_cmaps = ["rocket", "mako", "icefire", "vlag"]
     surfer_cmaps += [name + "_r" for name in surfer_cmaps]
--- a/surfer/utils.py
+++ b/surfer/utils.py
@@ -473,16 +473,16 @@ def create_color_lut(cmap, n_colors=256, center=None):
         Input colormap definition. This can be the name of a matplotlib
         colormap, a list of valid matplotlib colors, or a suitable
         mayavi LUT (possibly missing the alpha channel).
-        
-        if value is "auto", a default sequential or divergent colormap is 
+
+        if value is "auto", a default sequential or divergent colormap is
         returned
     n_colors : int, optional
         Number of colors in the resulting LUT. This is ignored if cmap
         is a 2d array.
     center : double, optional
-        indicates whether desired colormap should be for divergent values, 
+        indicates whether desired colormap should be for divergent values,
         currently only used to select default colormap for cmap='auto'
-        
+
     Returns
     -------
     lut : n_colors x 4 integer array
@@ -500,7 +500,7 @@ def create_color_lut(cmap, n_colors=256, center=None):
 
             return lut
 
-    # choose default colormaps (REMEMBER to change doc, e.g., in 
+    # choose default colormaps (REMEMBER to change doc, e.g., in
     # Brain.add_data, when changing these defaults)
     if isinstance(cmap, string_types) and cmap == "auto":
         if center is None:
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -959,15 +959,14 @@ class Brain(object):
         self.overlays_dict[name] = ol
         self._toggle_render(True, views)
 
-    	@verbose
-		def add_data(self, array, min=None, max=None, mid=None, 
-                 thresh=None, center=None, transparent=False, 
+    @verbose
+	def add_data(self, array, min=None, max=None, thresh=None, 
                  colormap="auto", alpha=1,
                  vertices=None, smoothing_steps=20, time=None,
                  time_label="time index=%d", colorbar=True,
                  hemi=None, remove_existing=False, time_label_size=14,
-                 initial_time=None, scale_factor=None, vector_alpha=None,
-                 verbose=None):
+                 initial_time=None, scale_factor=None, vector_alpha=None, 
+                 mid=None, center=None, transparent=False, verbose=None):
         """Display data from a numpy array on the surface.
 
         This provides a similar interface to
@@ -992,10 +991,10 @@ class Brain(object):
             and pass ``time_label=None``.
         min : float
             min value in colormap (uses real min if None)
-        max : float
-            max value in colormap (uses real max if None)
         mid : float
             intermediate value in colormap (middle between min and max if None)
+        max : float
+            max value in colormap (uses real max if None)
         thresh : None or float
             if not None, values below thresh will not be visible
         center : float or None
@@ -1008,8 +1007,8 @@ class Brain(object):
             name of matplotlib colormap to use, a list of matplotlib colors,
             or a custom look up table (an n x 4 array coded with RBGA values
             between 0 and 255), the default "auto" chooses a default divergent 
-            colormap, if "center" is given, otherwise a default sequential 
-            colormap.
+            colormap, if "center" is given (currently "icefire"), otherwise a 
+            default sequential colormap (currently "rocket").
         alpha : float in [0, 1]
             alpha level to control opacity of the overlay.
         vertices : numpy array
@@ -1042,8 +1041,6 @@ class Brain(object):
         vector_alpha : float | None
             alpha level to control opacity of the arrows. Only used for
             vector-valued data. If None (default), ``alpha`` is used.
-        verbose : bool, str, int, or None
-            If not None, override default verbose level (see surfer.verbose).
 
         Notes
         -----
@@ -1866,13 +1863,13 @@ class Brain(object):
         divergent indicate this by providing a value for 'center'. The 
         meanings of fmin, fmid and fmax are different for sequential and 
         divergent colormaps. For sequential colormaps the colormap is 
-        characterised by:
+        characterised by::
             
             [fmin, fmid, fmax]
         
         where fmin and fmax define the edges of the colormap and fmid will be
         the value mapped to the center of the originally chosen colormap. For
-        divergent colormaps the colormap is characterised by
+        divergent colormaps the colormap is characterised by::
             
             [center-fmax, center-fmid, center-fmin, center, 
              center+fmin, center+fmid, center+fmax]
@@ -1890,8 +1887,8 @@ class Brain(object):
         fmax : float
             maximum value for colormap
         transparent : boolean
-            if True: use a linear transparency between fmin and fmid, or 
-            between center-fmin to center-fmid and center+fmin to center+fmid
+            if True: use a linear transparency between fmin and fmid
+            (symmetrically for divergent colormaps)
         center : float
             if not None, gives the data value that should be mapped to the 
             center of the (divergent) colormap
@@ -2761,7 +2758,7 @@ def _scale_mayavi_lut(lut_table, fmin, fmid, fmax, transparent,
     """
     if not (fmin < fmid) and (fmid < fmax):
         raise ValueError("Invalid colormap, we need fmin<fmid<fmax")
-    if not (alpha >= 0) and (alpha <= 1):
+    if not 0 <= alpha <= 1:
         raise ValueError("Invalid alpha: it needs to be within [0, 1]")
 
     # Cast inputs to float to prevent integer division
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -2813,14 +2813,33 @@ def _scale_mayavi_lut(lut_table, fmin, fmid, fmax, transparent,
         n_fill = int(round(fmin * n_colors2 / (fmax-fmin))) * 2
         fillcol = lut_table[n_colors2, :].copy()
         fillcol[-1] = 0
-        return np.r_[
+        lut_table = np.r_[
                 _scale_sequential_lut(lut_table[:n_colors2, :], 
                                       center-fmax, center-fmid, center-fmin), 
                 np.tile(fillcol, (n_fill, 1)),
                 _scale_sequential_lut(lut_table[n_colors2:, :], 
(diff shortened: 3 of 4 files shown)


examples/plot_resting_correlations.py
```diff
@@ -46,14 +46,25 @@
...
 """
-brain.add_data(surf_data_lh, -.7, .7, colormap="vlag", alpha=.75,
-               hemi='lh')
-brain.add_data(surf_data_rh, -.7, .7, colormap="vlag", alpha=.75,
-               hemi='rh')
+You can make the small correlations transparent by re-scaling the colormap with
+transparency and letting the rescaler know that this is a divergent color map.
+In this case the first number gives the value that should be in the center of
+the colormap, the second determines which values will be transparent and the 
+third determines the left and right edges of the new colormap.
```
larsoner: It's also common to want thresholds for divergent colormaps to effectively be:
```
[-fmax, -fmid, -fmin, 0, fmin, fmid, fmax]
```
Where `[-fmin, fmin]` is a single color (the midpoint of the colormap).

If we support this use case somehow, your use case is also automatically supported. Can you think of a good API?

One idea would be to do `center=None` default (for non-divergent) and do:

1. `add_data(...., fmin=2, fmid=4, fmax=6, center=0)` would do what I want
2. `add_data(..., fmin=0, fmid=3, fmax=6, center=0)` would do what you want

Or something equivalent and maybe nicer could be done with `divergent=False` default:

1. `add_data(..., fmin=2, fmid=4, fmax=6, divergent=True)` would do what I want
2. `add_data(..., fmin=0, fmid=3, fmax=6, divergent=True)` would do what you want

WDYT?

mwaskom: I like `center=` as a way of specifying divergent colormaps.

mwaskom: It would be nice if pysurfer did what seaborn does with divergent colormaps which is to clip the lookup table so that the colorbar shows the extreme values in the data while keeping the luminance midpoint in the right place.

mwaskom: Also if @larsoner's original suggestion is implemented that would mean completely duplicating the `add_overlay` functionality, which is worth thinking about.

sbitzer: Yes, originally I also planned to allow something like `[-fmax, -fmid, -fmin, 0, fmin, fmid, fmax]`, but settled for the easier `[c-max, c-mid, c, c+mid, c+max]`, because that's much easier to handle with a single colormap. At the latest when thinking about generating a colorbar for the version allowing for space between the positive and negative ranges I realised that this is in effect plotting two separate sets of data with two separate colormaps. I only then found add_overlay which, when called as in the example, already does what you want, just that it doesn't allow you to specify colormaps (yet?). 

As a user, I'm not actually clear about what the principle difference between add_overlay and add_data is. It appears to me that add_overlay is just a crippled version of add_data?

I also like the `center=` argument, but it makes less sense, when there is no separation between negative and positive sides, because you'd have to give both `fmin` and `center` values which should be equal. `center` could also override `fmin`, when given, which sounds fine, too.

@mwaskom I'm afraid I didn't quite understand what you mean with seaborn's handling of diverging colormaps. Intuitively, the procedure you're describing sounds like what the fmax-argument already does.

mwaskom: @sbitzer described more here: https://github.com/mwaskom/seaborn/pull/1184. It's a little related to your core aim (intermediate transparency) but as I've thought about adding `center` to PySurfer to handle this aspect of diverging colormaps it might be good to do all at once.

mwaskom: > As a user, I'm not actually clear about what the principle difference between add_overlay and add_data is. It appears to me that add_overlay is just a crippled version of add_data?

`add_overlay` existed first to closely reproduce tksurfer functionality. `add_data` was later added to provide a more direct mapping from an array to a heatmap overlay and gradually grew increasingly bloated with additional functionality. My preference has long been to break it up.

larsoner: `add_data` does a lot more than `add_overlay`, I also view it as having (almost) a superset of functionality.

Not sure about breaking it up, sounds like a lot of work for little gain but if someone is motivated (and can keep it working well) then it's fine by me.

I don't think we should go with just `[c-max, c-min, c, c+min, c+max]` it is less functional.

mwaskom: > I don't think we should go with just [c-max, c-min, c, c+min, c+max] it is less functional.

That's not my proposal.

larsoner: That was meant as a response to @sbitzer's comment, sorry if it was ambiguous

larsoner: ... but seeing as in `seaborn` the problem is a bit different than what we have here (and what I have proposed with `[-fmax, -fmid, -fmin, 0, fmin, fmid, fmax]` or a variant like `center + [...]`), can you elaborate on the proposal you'd prefer here? Would it be able to get the `[-fmax, -fmid, -fmin, 0, fmin, fmid, fmax]` functionality?

mwaskom: @larsoner yes the seaborn-relevant point I'm making is just that it would be nice if diverging colormaps could satisfy both (1) there is a specified "center value" and (2) the negative and positive colorbar limits reflect the limits of the data. I understand the discussion above about whether the parameterization should be a boolean `diverging` with an implicit `center=0` or an explicit `center` parameter.

larsoner: > it would be nice if diverging colormaps could satisfy both (1) there is a specified "center value" and (2) the negative and positive colorbar limits reflect the limits of the data. 

My proposal only does (1) but not (2) IIUC. By (2) you mean supporting something like (in an extreme case) `[-2, -1.5, -1, 0, 10, 11, 12]`, right? The only way I can see to get this sort of behavior (without too much complexity) is just to allow people to specify 7 control points if they want.

mwaskom: @larsoner seaborn has three parameters: `vmin`, `vmax`, and `center`. To get an opacity ramp you'd need a fourth (`vmid`).

larsoner: I'm not sure how `vmid` would work -- where would it go in terms of the control points / what would the equivalent be in the 7-point scheme?

larsoner: (My understanding of the seaborn scheme is that it would be equivalent to the seven-point scheme `[vmin, (vmin+center)/2., center, center, center, (vmax+center)/2., vmax]`)

mwaskom: Here's a concrete example. You want to plot an overlay that ranges from -5 to 8. You put `add_data(..., center=0, fmin=2, fmid=4, fmax=8)`. The colorbar that you see has limits at -5 and 8. It is fully opaque in the ranges [-5, -4] and [4, 8]. It ramps down in opacity linearly between [-4, -2] and ramps up in opacity linearly between [2, 4]. It is transparent in the range [-2, 2]. (All range specs ignore details of inclusive/exclusive for now). The middle RGB value of the colormap (LUT128) corresponds to 0, the maximum RGB value of the colormap (LUT255) corresponds to 8, and an intermediate RGB value of the colormap (LUT48) corresponds to -5.

larsoner: Yeah that's fine by me -- it's equivalent to `[-8, -4, -2, 0, 2, 4, 8]` in the seven-point scheme I keep talking about -- and AFAICT already equivalent to what I proposed yesterday (right?).

However, you mentioned also wanting to have support for divergent colormaps where "(2) the negative and positive colorbar limits reflect the limits of the data.", how in this `center,fmin,fmid,fmax` 4-kwarg scheme do we get that? Maybe I don't understand what you mean by it...?

larsoner: (I assume that by (2) you mean that you want LUT0 to correspond to `vmin` and LUT255 to correspond to `vmax` in seaborn terminology, but we don't have independent/equivalents for `vmin`/`vmax` in this scheme because we force symmetry as `vmin=-fmax, vmax=fmax` effectively)

mwaskom: I'll just drop this for now and code it up later if I become interested again.

larsoner: So I think the existing proposal is to use `center, fmin, fmid, fmax` as:

1. Default `center=None`, treat as sequential colormap (current behavior)
2. `center is not None` use `[center-fmax, center-fmid, center-fmin, center, fmin, fmid, fmax]` as limits (with center-fmin to fmin as LUT128, and -fmax/fmax as LUT0 and LUT255, respectively), with appropriate checks for monotonic `center <= fmin <= fmid <= fmax`

Okay with you to proceed with this @mwaskom?

mwaskom: Seems fine.

larsoner: @sbitzer does this make sense to you? Do you have time to give it a shot?

sbitzer: Yes, makes sense, except that I guess you mean `[center-fmax, center-fmid, center-fmin, center, center+fmin, center+fmid, center+fmax]`?

I'll see how far I get today or tomorrow and let you know of the progress.

larsoner: Yes, I think so. Thanks for taking a stab at it. A couple of minor points:

- It might be worth de-duplicating with `add_overlay` if/where possible.
- We probably want different default colormaps to be used (`cmap='auto'` by default?) depending on if `center=None` (use current default) or not (some diverging colormap). 
- It would be good if the colormap actually displays properly on the colorbar. It might, depending on how it's implemented, but we should at least check the result for e.g. `center=0, fmin=1, fmid=2, fmax=3`

@mwaskom any personal favorite diverging colormap we should use by default?

mwaskom: @larsoner the `icefire` colormap I added recently is intended to be used for diverging neuroimaging visualization.

sbitzer: Ok, automatically choosing a divergent colormap, when `center` is provided is a good idea. The only issue is that so far the colormap is only chosen in `add_data` while `center` would live in `scale_data_colormap`. Should I then also add the full `fmin, fmid, fmax, center` to `add_data`?

larsoner: It's okay with me, as they'll mostly get passed through (and it could save a call to `scale_data_colormap`). Would it simplify / unify `add_data` and `add_overlay` calls a bit?

sbitzer: I did most of the work today, but still see a strange effect that I don't understand. I will be away the next two days and will check in my progress on Friday.

sbitzer:
Ok, these changes should implement the discussed functionality. I had some issues with Mayavi/VTK when changing the number of colors in the lookup table. So I now ensure that the new lookup tables have exactly 256 colors (see 59debb4). I'm most uncertain about the compatibility of the newly added arguments in `add_data` with existing code. Maybe you would like to comment on that. I have not further thought about dissociation of `add_data` and `add_overlay`.

surfer/utils.py
```diff
@@ -473,9 +473,15 @@ def create_color_lut(cmap, n_colors=256):
...
         mayavi LUT (possibly missing the alpha channel).
+        
+        if value is "auto", a default sequential or divergent colormap is 
+        returned
     n_colors : int, optional
         Number of colors in the resulting LUT. This is ignored if cmap
         is a 2d array.
+    center : double, optional
+        indicates whether desired colormap should be for divergent values, 
+        currently only used to select default colormap for cmap='auto'
```
larsoner: missing newline

surfer/utils.py
```diff
@@ -493,6 +499,12 @@ def create_color_lut(cmap, n_colors=256):
 
             return lut
 
+    if isinstance(cmap, string_types) and cmap=="auto":
```
larsoner: spaces around operator (these are PEP8 things)

surfer/viz.py
```diff
@@ -960,8 +960,9 @@ def add_overlay(self, source, min=2, max="robust_max", sign="abs",
         self.overlays_dict[name] = ol
         self._toggle_render(True, views)
 
-    def add_data(self, array, min=None, max=None, thresh=None,
-                 colormap="RdBu_r", alpha=1,
+    def add_data(self, array, min=None, max=None, mid=None, 
+                 thresh=None, center=None, transparent=False,
```
larsoner: even though it's annoying, we probably need to put `mid`, `center`, and `transparent` at the end of the argument list for backward compatibility

surfer/viz.py
```diff
@@ -992,12 +993,22 @@ def add_data(self, array, min=None, max=None, thresh=None,
...
+            min, max and mid, see ``scale_data_colormap`` for further info
+        transparent : bool
+            whether colors from mid to min should become increasingly 
+            transparent
         colormap : string, list of colors, or array
             name of matplotlib colormap to use, a list of matplotlib colors,
             or a custom look up table (an n x 4 array coded with RBGA values
-            between 0 and 255).
+            between 0 and 255), the default "auto" chooses a default divergent 
+            colormap, if "center" is given, otherwise a default sequential
```
larsoner: might as well list what they will be, e.g. '... default divergent colormap ("icefire") if "center" is given, otherwise ...'

surfer/viz.py
```diff
@@ -1833,30 +1854,59 @@ def set_surf(self, surf):
...
-    def scale_data_colormap(self, fmin, fmid, fmax, transparent, verbose=None):
+    def scale_data_colormap(self, fmin, fmid, fmax, transparent, 
+                            center=None, alpha=1.0, verbose=None):
         """Scale the data colormap.
+        
+        The colormap may be sequential or divergent. When the colormap is 
+        divergent indicate this by providing a value for 'center'. The 
+        meanings of fmin, fmid and fmax are different for sequential and 
+        divergent colormaps. For sequential colormaps the colormap is 
+        characterised by:
```
larsoner: `::` here, it will render what follows nicely as code

surfer/viz.py
```diff
@@ -1833,30 +1854,59 @@ def set_surf(self, surf):
...
+            
+            [fmin, fmid, fmax]
+        
+        where fmin and fmax define the edges of the colormap and fmid will be
+        the value mapped to the center of the originally chosen colormap. For
+        divergent colormaps the colormap is characterised by
```
larsoner: ::

surfer/viz.py
```diff
@@ -1833,30 +1854,59 @@ def set_surf(self, surf):
...
-            minimum value of colormap
+            minimum value for colormap
         fmid : float
             value corresponding to color midpoint
         fmax : float
             maximum value for colormap
         transparent : boolean
-            if True: use a linear transparency between fmin and fmid
+            if True: use a linear transparency between fmin and fmid, or 
+            between center-fmin to center-fmid and center+fmin to center+fmid
```
larsoner: How about "... linear transparency between fmin and fmid (symmetrically for divergent colormaps)."

surfer/viz.py
```diff
@@ -2662,44 +2757,85 @@ def _scale_mayavi_lut(lut_table, fmin, fmid, fmax, transparent,
     """
     if not (fmin < fmid) and (fmid < fmax):
         raise ValueError("Invalid colormap, we need fmin<fmid<fmax")
+    if not (alpha >= 0) and (alpha <= 1):
```
larsoner: `if not 0 <= alpha <= 1`

surfer/viz.py
```diff
@@ -2662,44 +2757,85 @@ def _scale_mayavi_lut(lut_table, fmin, fmid, fmax, transparent,
...
     fmin = float(fmin)
     fmid = float(fmid)
     fmax = float(fmax)
 
-    logger.info("colormap: fmin=%0.2e fmid=%0.2e fmax=%0.2e "
-                "transparent=%d" % (fmin, fmid, fmax, transparent))
+    divergent = center is not None
+
+    if divergent:
+        logger.info("colormap: fmin=%0.2e fmid=%0.2e fmax=%0.2e "
```
larsoner: How about messages formatted more like this (aiming to be compact + readable):
```
colormap divergent: center=0., [2., 3., 4.] (transparent)
colormap sequential: [0, 5, 10] (opaque)
```

larsoner:
Also, could you rebase? That way the CIs can check everything.

surfer/viz.py
mwaskom: One reason not to do that is that encoding defaults in text that doesn't automatically update with the code is asking for trouble later if things change.

larsoner: That's true. There's a tradeoff to be made there. In this case, once we choose defaults it's backward-incompatible to change them, so the change is likely to be invasive enough we'd notice that we need to update the defaults. I think the informativeness to the user is worth it in this case -- what do you think?

mwaskom: In this case, breaking backwards compatibility wouldn't cause any existing code to stop working or produce incorrect results, it would simply look different. So in the future, if we decide there is a better default colormap to use, I would feel no compunction about the change.

mwaskom: (That said, the worst thing that would happen if the docs don't get updated is some minor confusion for an already curious user, so maybe it's fine).

larsoner: >  wouldn't cause any existing code to stop working or produce incorrect results, it would simply look different. 

For scripts that use PySurfer to make figures for publication (hopefully there are many of these!) this would be problematic, so I wouldn't want to do this without a deprecation cycle.

mwaskom: Well, you only publish once, and it will be extremely obvious if/when it changes, so that doesn't strike me as a problem. With this sort of parameter my opinion is that not specifying it is "asking for the default", which may be subject to change. In other words, best practices for publication scrips that you really don't want to change is to encode the values you really want even when they correspond to the default at the time you write them.

Overly stubborn resistance to changing default colormap even when you know it's bad is what caused the sourge of jet after all :)

Anyway it's an orthogonal issue and I'm fine saying in the docs what the default is. Maybe add a comment in the code that the docstring should be updated if the defaults are changed.

surfer/viz.py
sbitzer: but is there a convention that you have been using for these messages, or does anything depend on their previous format?

sbitzer:
I have resolved the few things you discussed except for the log message (see my comment above). The colorbar in your example looks strange, because I used the first color of the second half of the colormap to fill up the center of the colormap. This works nicely with most divergent colormaps as they smoothly transition from one to the other side of the colormap. That's not the case for the colormap in your example. I have now included some code which tries to identify these colormaps and then chooses gray as the fill color in the center for them.

A practical issue as I haven't done that before: I have rebased my branch with nipy/master locally and my state is now:
```On branch divtrans
Your branch and 'origin/divtrans' have diverged,
and have 17 and 7 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)```

I can't imagine that merging the old commits is the appropriate thing to do. What is?

File: surfer/viz.py (hunk 10 above)
Comment (larsoner) [reply surfer/viz.py]: I think it's okay to change it to make it clearer / more compact, it is very unlikely someone's code relies on parsing the colormap log message

**Comment by larsoner** [comment]
> I can't imagine that merging the old commits is the appropriate thing to do. What is?

I would do:
```
git checkout -b divtrans-bak origin/divtrans  # make a backup
git checkout divtrans
git push origin --force divtrans
```

> The colorbar in your example looks strange, because I used the first color of the second half of the colormap to fill up the center of the colormap. This works nicely with most divergent colormaps as they smoothly transition from one to the other side of the colormap. I have now included some code which tries to identify these colormaps and then chooses gray as the fill color in the center for them.

The problem is actually more general than this. Gray should be used anytime that `transparent=True` in the `[-fmin, fmin]` region, as the data values in that span will be gray. (Another way to say this is that the `mayavi` colorbar function does not implement transparency correctly in this case / wouldn't know what to put underneath the colorbar overlay to make transparency look correct.) Make sense?

Ideally the `-fmid, -fmin` and `fmin, fmid` regions with `transparent=True` would show up on the colorbar as linear mixes between gray and the actual colors, too. But I'm not sure if you can modify the LUT for the colorbar display and the LUT for the data separately, which seems like would be required to get this behavior. (Alternatively, if there is some way to get the colorbar display to actually show transparency, and pick a background color to blend with, all of this would become much simpler.)

**Comment by sbitzer** [comment]
Yes, I agree that it would make most sense, if the colorbar would somehow reflect transparency properly, but I'm afraid that this is a Mayavi or even VTK issue that I currently don't want to go into. At least for `transparent=True` I find the workaround of extending the region for the middle color quite intuitive, because it's clear then that the middle color is fully transparent. I have also seen in a talk last week a brain visualisation using this (don't know which software they used, but their colorbar was almost all black with some color gradients at the edges and the brain showed selected blobs). Only for `transparent=False` with `fmin>0` confusion may occur, because you won't be able to see a representation of `fmin` on the colorbar although you might expect a sharp edge. Hm, to prevent this we could simply choose gray as center fill color whenever `transparent=False`? I still find it visually more pleasing to fill in the center color from the colormap for `transparent=True`, though.

**Comment by larsoner** [comment]
> Yes, I agree that it would make most sense, if the colorbar would somehow reflect transparency properly, but I'm afraid that this is a Mayavi or even VTK issue that I currently don't want to go into

It looks like the `mayavi.mlab.colorbar` has an `object` property that allows you to choose a color. So what we could do is:

1. Create two colormaps: one that is actually used for the data (you have this already) and
2. Another that is that same colormap mixed with a `(0.5, 0.5, 0.5)` background.
3. Apply the first colormap to the data (you have done this already), and
4. Apply the second colormap to a dummy object (single-triangle mesh at the origin of tiny size)
5. Set the dummy object visibility off
6. Call `colormap` with `object=dummy`

It's a bit of a hack but it should work, and I suspect will best represent what's happening with the data. I can give it a try if you're tired of messing with Mayavi :)

>  I find the workaround of extending the region for the middle color quite intuitive, because it's clear then that the middle color is fully transparent.

Hmmm... I'm not sure I agree that the transparent effect is currently made clear by the colorbar. With `transparent=True`, those middle colors are `alpha=0.`, i.e., invisible / gray (on a standard brain), but show up as whatever the middle color is on the colorbar.

**Comment by sbitzer** [comment]
> It's a bit of a hack but it should work, and I suspect will best represent what's happening with the data. I can give it a try if you're tired of messing with Mayavi :)
It sounds like it could work. Sure, feel free to see how it goes!

I wondered more generally why Mayavi/VTK programmers did not include the possibility of having transparent colorbars and figured out that this is not true: You can switch on proper transparency of the colorbar with
```
surf = brain.data_dict['lh']['surfaces'][0]
lm = surf.module_manager.scalar_lut_manager
lm.scalar_bar.use_opacity = 1
lm.render()
```
This will apply transparency with respect to the figure window background (white in my example). For `transparent=False` this gives a result equivalent to `add_overlay`:

image

but with `transparent=True` and `alpha=0.75` I'm not sure whether the result is a good representation of the colors on the brain:

image

I guess the question is what the background colour should be against which the transparency is plotted. Given that the colours apply to the brain, this should probably, in the ideal case be a texture containing the two colours of the inflated brain (greys in most cases). I wonder whether people would intuitively recognise a simple mixture between the colorbar and a particular grey to represent the transparency on the brain. Perhaps the transparency of the colorbar with the figure background is already a good approximation?

**Comment by larsoner** [comment]
> For transparent=False this gives a result equivalent to add_overlay:

In this case I wouldn't have expected to be able to see any brain, i.e. the `-fmin, fmin` interval should have been black (or really `-fmin, 0` the "just left of colormap midpoint" color and `0, fmin` to be "just right of colormap midpoint, which in this case are both black).

> but with transparent=True and alpha=0.75 I'm not sure whether the result is a good representation of the colors on the brain:

It looks better than before, but I agree still not quite right.

> I guess the question is what the background colour should be against which the transparency is plotted. Given that the colours apply to the brain, this should probably, in the ideal case be a texture containing the two colours of the inflated brain (greys in most cases).

Yes this would be ideal, but I think just using middle gray `(0.5, 0.5, 0.5)` would already take care of 95% plus of use cases, and easy to describe in the docs (i.e., say we blend with this).

> I wonder whether people would intuitively recognise a simple mixture between the colorbar and a particular grey to represent the transparency on the brain.

I think they would. In every example you've posted so far, and even in the examples we have online (even with the `bone`-colored brain) I think it would look okay, and probably more representative than anything else so far.

> Perhaps the transparency of the colorbar with the figure background is already a good approximation?

I'd say it's better than what we have in `master` but not as good as mixing with gray (by default) would be for our specific use case. If people like mixing with the background instead of gray we could add an argument to `scale_data_colormap` to also allow that mode. But maybe we won't need it once we see the gray solution.

Did you have a chance to try my suggestion of using a dummy object for comparison? (Or alternatively, to see if there is a way to set the background color of the colorbar in addition to setting the opacity?)

**Comment by sbitzer** [comment]
I did not find an option to set the background only of the colorbar itself, but this is possible:

image

Not sure whether I like the box or whether the grey is better than the white. I did not try your dummy object approach.

**Comment by larsoner** [comment]
I think the box looks a bit worse than no box.

And can you look into this problem?

> > For transparent=False this gives a result equivalent to add_overlay:

> In this case I wouldn't have expected to be able to see any brain, i.e. the -fmin, fmin interval should have been black (or really -fmin, 0 the "just left of colormap midpoint" color and 0, fmin to be "just right of colormap midpoint, which in this case are both black).

**Comment by sbitzer** [comment]
Ah right, I read this over. I interpreted `fmin` as the first value that you want to see displayed while `transparent` only controls whether there is a transparency gradient between `fmin` and `fmid`, but you're right, the `add_data`-interface defines this differently for sequential colormaps. I can fix that. Do you perhaps know how I can change the call
```
pipe = threshold_filter(pipe, low=thresh, figure=self._f)
```
to one in which all values except those in `[center - thresh, center + thresh]` are kept?

larsoner:
Do we even need the thresholding filter? Maybe setting the limits properly (and scaling our colorbar properly) will already work.

sbitzer:
I don't see how setting the limits and scaling the colorbar can implement the selective clipping of values at the centre of the colormap. The colormap used by Mayavi will always linearly interpolate colours given scalar values so you cannot simply skip values that lie in the centre. As I understood it, the `thresh` argument allows that kind of clipping of 'small' values independent of the transparency of the colormap and I had previously interpreted `fmin` to behave like that. As you pointed out, however, `fmin` is supposed to be the value from which all equal or smaller magnitude values are mapped to the leftmost colour in sequential colormaps or the centre colour(s) in divergent colormaps. So I don't see how the functionality of `thresh` could be implemented with colormap properties. Do I miss something?

The simplest solution would be to not adapt the behaviour of `thresh` for divergent colormaps, noting that with `transparent=True` a similar visualisation can be achieved.

larsoner:
> I don't see how setting the limits and scaling the colorbar can implement the selective clipping of values at the centre of the colormap.

My point is just that I don't think these values should be / need to be clipped, at the very least in the `transparent=True` case and maybe also in the `transparent=False` case.

sbitzer:
Ok, I agree and, therefore, didn't touch the `thresh` functionality. I have corrected the behaviour of `fmin` though, updated the log message and introduced transparency of the colorbar with respect to the figure background in the latest commits.

larsoner:
Looks like there are some PEP8 errors:

https://travis-ci.org/nipy/PySurfer/jobs/284137743#L1221

Can you rebase so that CircleCI runs the doc build properly?

larsoner:
See https://github.com/sbitzer/PySurfer/pull/1 for non-hack gray fix

larsoner:
Looks nice:

https://27-1551431-gh.circle-artifacts.com/0/tmp/circle-artifacts.YB9nYb3/html/auto_examples/plot_meg_inverse_solution.html

But something strange has happened here:

https://27-1551431-gh.circle-artifacts.com/0/tmp/circle-artifacts.YB9nYb3/html/auto_examples/plot_vector_meg_inverse_solution.html#sphx-glr-auto-examples-plot-vector-meg-inverse-solution-py

(it's probably a bug in the code I wrote, feel free to look or I can do it)

sbitzer:
Nice!

> (it's probably a bug in the code I wrote, feel free to look or I can do it)
As I wouldn't know where to start, except for the obvious that it has something to do with the computation of the background color, I would appreciate, if you could check it out first.

larsoner:
> I would appreciate, if you could check it out first.

I should be able to take a look tomorrow

larsoner:
Fix up again at https://github.com/sbitzer/PySurfer/pull/1

sbitzer:
Great, fine by me! :+1:

larsoner:
Anybody else want to look? If not I'll go ahead and merge.

larsoner:
Thanks @sbitzer !

PR #208, opened 2017-09-25
MRG: Add verbose argument

larsoner:
We get a ton of these messages in MNE because we use `add_data`:
```
Updating smoothing matrix, be patient..
Smoothing matrix creation, step 1
Smoothing matrix creation, step 2
Smoothing matrix creation, step 3
Smoothing matrix creation, step 4
Smoothing matrix creation, step 5
Smoothing matrix creation, step 6
Smoothing matrix creation, step 7
Smoothing matrix creation, step 8
Smoothing matrix creation, step 9
Smoothing matrix creation, step 10
```
This will allow us to pass `verbose=False` to avoid them (this process is quick enough we don't need to be informed of the progress).

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -960,12 +960,14 @@ class Brain(object):
         self.overlays_dict[name] = ol
         self._toggle_render(True, views)
 
+    @verbose
     def add_data(self, array, min=None, max=None, thresh=None,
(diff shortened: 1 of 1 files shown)


larsoner:
Travis failures are unrelated and fixed by #205

PR #210, opened 2017-09-29
MRG: Add CircleCI

larsoner:
Let's see if this works.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -43,7 +43,7 @@ dependencies:
       else
         echo "Conda env already set up.";
       fi;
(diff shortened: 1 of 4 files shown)


larsoner:
Okay this gives us easier doc-checking CI:

https://14-1551431-gh.circle-artifacts.com/0/tmp/circle-artifacts.ZhSjlcj/html/auto_examples/index.html

About 1/3 of examples don't run because the `fsaverage_min.zip` I made for testing doesn't provide all necessary surfaces. We can change that later if someone feels like updating the dataset.

Ready for review/merge from my end.

PR #211, opened 2017-10-02
MRG: Always force GUI updates

larsoner:
@nfoti can you see if this fixes your problem?

This forces a GUI process loop update when calling `_force_render`, which is useful e.g. to deal with the toolbar removal at the beginning. @nfoti found that getting the model-view matrix in a script was different before/after allowing the interpreter to idle, which is a sign that the idle-GUI-loop code was doing something to the display, which shouldn't happen if no interaction has taken place and we are properly forcing rendering changes.

Tests pass locally, let's see if the CIs are happy.

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -1,3 +1,4 @@
+import logging
 from math import floor
 import os
 from os.path import join as pjoin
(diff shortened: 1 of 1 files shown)


larsoner:
Examples still look okay:

https://16-1551431-gh.circle-artifacts.com/0/tmp/circle-artifacts.dQ84wpM/html/auto_examples/index.html

nfoti:
@larsoner This fixed my problem.

Thanks.

larsoner:
Great, ready for review/merge from my end, then

agramfort:
thx @larsoner

PR #212, opened 2017-10-03
Fix annotation colors

wmvanvliet:
When annotating the brain with a brain atlas, the colors are not always set right:

```python
import surfer
brain = surfer.Brain('fsaverage', hemi='both', surf='inflated')
brain.add_annotation('aparc', borders=False)
```
fig1

This has to do with the reordering of `cmap` and the translation of the `ids` vector. This PR fixes it:

fig2


larsoner:
Thanks @wmvanvliet

PR #215, opened 2017-10-13
MRG: Fix text label

larsoner:
Fixes this problem:

https://larsoner.github.io/mne-biomag-group-demo/_images/sphx_glr_plot_source_stats_001.png

While I was at it, I ~~hit a couple of minor snags on Py3k so I added a 3.6 CI (via `environment.yml`) and also~~ moved over to `pytest` (`nose` is no longer maintained).


larsoner:
Okay I removed the Py3k CI because I couldn't get it to be stable, but the rest of this (including the transition to `pytest` from `nose`) is good to go and CIs are happy

https://41-1551431-gh.circle-artifacts.com/0/tmp/circle-artifacts.SmD6e3t/html/auto_examples/plot_meg_inverse_solution.html

agramfort:
LGTM

mwaskom:
Great! pytest is such an improvement.

PR #220, opened 2017-11-22
BUG: Add 'center' as a data property

wmvanvliet:
Fixes an error I get when opening a `TimeViewer` complaining about `'center'` missing from the data properties dict.


larsoner:
Thanks @wmvanvliet

PR #221, opened 2017-11-27
MRG: Fix imsave

### LINKED ISSUE #209: save_image broken
The `save_image` method tries to use the `imsave` function in `scipy.misc`. However, it appears that this function was removed from `scipy` some time ago (around 0.13). The `imsave` function now resides in `matplotlib.pyplot`.

You can recreate this bug with the following:
```
conda create -n mayavi python=3
source activate mayavi
conda install -c conda-forge mayavi scipy matplotlib
conda install ipython pyqt=4.11
pip install pysurfer
```

followed by:
```
ipython
$ %gui qt
$ from surfer import Brain
$ brain = Brain('fsaverage', 'split', ' 'inflated')
$ brain.save_image("./test.pdf")
```
(there's not really a need to do this in `ipython)

larsoner:
Closes #209.

@mwaskom I used your approach but left the figure-generation private for now. Feel free to make a follow-up PR to make it public with proper documentation, etc.

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -2260,6 +2260,16 @@ class Brain(object):
         if mlab.options.backend != 'test':
             mlab.savefig(filename, figure=brain._f)
 
+    def _screenshot_figure(self, mode='rgb', antialiased=False, dpi=100):
+        """Create a matplolib figure from the current screenshot."""
+        # adapted from matplotlib.image.imsave
+        from matplotlib.backends.backend_agg import FigureCanvasAgg
+        from matplotlib.figure import Figure
+        fig = Figure(dpi=dpi, frameon=False)  # DPI only used for metadata
+        FigureCanvasAgg(fig)
+        fig.figimage(self.screenshot(mode, antialiased), resize=True)
+        return fig
+
     def save_image(self, filename, mode='rgb', antialiased=False):
         """Save view from all panels to disk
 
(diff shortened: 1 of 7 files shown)


mwaskom:
Haven't tested but LGTM

surfer/viz.py
```diff
@@ -2652,7 +2652,7 @@ def save_movie(self, fname, time_dilation=4., tmin=None, tmax=None,
             raise ValueError("tmax=%r is greater than the latest time point "
                              "(%r)" % (tmax, self._times[-1]))
         n_frames = floor((tmax - tmin) * time_dilation * framerate)
-        times = np.arange(n_frames)
+        times = np.arange(n_frames).astype(float)
```
mwaskom: Probably not a big performance hit here, but you could pass the dtype to arange to avoid making a copy.

larsoner:
Okay CIs are happy (except pending OSX but that usually takes forever). I added a 3.6 CI, fixed related errors, and fixed a couple of bare `except:`s we had. Ready for review/merge from my end.

larsoner:
(actually not sure about AppVeyor yet, either, but it passed last time I let it finish)

christianbrodbeck:
Is there an advantage over using Matplotlib's `imsave` (https://github.com/matplotlib/matplotlib/blob/5f63dd42ec6deec5c85ab975d7d879c7c81cb161/lib/matplotlib/image.py#L1305), which seems to be more efficient for PNG but do the same otherwise?

larsoner:
I assumed it was just an efficiency thing so I didn't include it to reduce complexity, but I don't know for sure

christianbrodbeck:
Oh, I didn't mean to include the special case, but whether it wouldn't be easier to just use `matplotlib.image.imsave`

larsoner:
This moves toward a use case @mwaskom wanted, see his comments toward the end of #209

christianbrodbeck:
Aah, makes sense

larsoner:
Anybody else want to look (or @mwaskom want to look again now that everything passes)? If not I'll merge at the end of the week

PR #222, opened 2017-12-02
[MRG] View persistence

christianbrodbeck:
The problem can be seen by using a view in which roll is non-default, for example "ventral", and adding data; before this patch, the view changed after the add_data call.

Also adding a comment to docs to clarify something that confused me the first time I read it.

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -547,7 +547,7 @@ class Brain(object):
                 continue
 
             if state is False and view is None:
-                views[vi] = (mlab.view(figure=_f),
+                views[vi] = (mlab.view(figure=_f), mlab.roll(figure=_f),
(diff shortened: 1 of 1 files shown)


surfer/viz.py
```diff
@@ -556,7 +556,8 @@ def _toggle_render(self, state, views=None):
                 mlab.draw(figure=_f)
                 with warnings.catch_warnings(record=True):  # traits focalpoint
                     mlab.view(*view[0], figure=_f)
```
larsoner: `mlab.view` has a `roll` argument, could it be passed here instead of making a separate call?

christianbrodbeck: Yes seems to work too.

larsoner:
Thanks @christianbrodbeck

PR #223, opened 2017-12-04
FIX: Fix OSX

larsoner:
This will hopefully fix our OSX timeouts.


larsoner:
Indeed it did!

mwaskom:
So weird to see a green merge button for pysurfer :)

larsoner:
Hopefully it will stay that way for longer than a week or two this time :)

PR #226, opened 2018-02-06
MRG: Fix floating buttons

### LINKED ISSUE #225: Embedded screenshots in docs are very large
Was just looking at the docs and the screenshots now take up considerable more space than I remember.

For example: https://pysurfer.github.io/auto_examples/plot_fmri_activation.html

Here's how it looks to me with my Chrome set to 100% zoom:

screenshot 2018-02-03 15 08 15

Doesn't seem limited to that example.

Did this change when the docs moved to sphinx-gallery? Can the size of the embedded image be controlled (as I assume it used to be?) Having the figures spill over into the sidebar looks messy.

larsoner:
Closes #225.


larsoner:
@mwaskom scaling fixed:

https://81-1551431-gh.circle-artifacts.com/0/tmp/circle-artifacts.59wkDE6/html/auto_examples/plot_basics.html#sphx-glr-auto-examples-plot-basics-py

mwaskom:
LGTM. Full size is still a little big, but the responsive scaling is nice and I don't think there's need to fuss over it.

PR #232, opened 2018-03-29
FIX: Fix set surf

larsoner:


larsoner:
@christianbrodbeck can you confirm this fixes the bug you pointed out in https://github.com/mne-tools/mne-python/pull/5012 ?

christianbrodbeck:
Yes this fixes it. But it's not necessary for older versions. Must be that the dev version of mayavi makes a copy where the earlier version used a view (hence in-place update of the array in `load_geometry()`).

larsoner:
Ahh -- it would actually make sense for them to do this, it should be faster to only upload things that are known to have changed on update, which you can't for in-place array changes. It might have been luck that it worked before. That is my suspicion anyway based on my limitied graphics optimization knowledge. But feel free to open a Mayavi bug if you're curious.

christianbrodbeck:
Yeah that's plausible that that's behind the change. Glad you caught it so quickly!

PR #235, opened 2018-04-18
MRG: CircleCI 2.0

larsoner:
Migrate to CircleCI 2.0 as 1.0 reaches EOL in August.


larsoner:
Looks like we are good to go here:

https://96-1551431-gh.circle-artifacts.com/0/html/index.html

mwaskom:
Thanks @larsoner

PR #239, opened 2018-05-28
ADD: first draft of plot_custom_colours script

### LINKED ISSUE #228: Q: plotting [source x RGBA] array on brain surface
Hi there,

I am trying to plot an array of RGBA values (shape: [source x RGBA], [5124 x 4]) on a pysurfer brain. In other words, I want to manually specify the RGBA value at each vertex. The brain object can be from MNE-Python or Pysurfer natively, I don't mind. I have tried playing around with the LUT but I think I am making things much more complicated than they need to be. 

Do you know of a way I could achieve this? Perhaps one method would be to access the RGBA values of the brain image itself, and simply add my RGBA values on top of it; but I can't find the cortex RGBA values to manipulate, and maybe you know a better method.

It's been suggested to me that @larsoner may be one of the people to ask - I hope you don't mind me flagging you here.

Thank you very much in advance. I really appreciate it!
Laura

LauraGwilliams:
Hi @larsoner,

I have (finally - I'm sorry) put together a draft of the script that allows for plotting custom rgba values. Any feedback is appreciated!

Thanks,
Laura

Closes #228

### DIFF
--- a/examples/plot_custom_colors.py
+++ b/examples/plot_custom_colors.py
@@ -26,7 +26,7 @@ def norm(x):
 # 2) init brain and get spatial co-ordinates
 
 # params
-# subjects_dir = os.environ['SUBJECTS_DIR']
+subjects_dir = os.environ['SUBJECTS_DIR']
 hemi = 'lh'
 surf = 'white'
 
--- a/examples/plot_custom_colors.py
+++ b/examples/plot_custom_colors.py
@@ -26,14 +26,14 @@ def norm(x):
 # 2) init brain and get spatial co-ordinates
 
 # params
-subjects_dir = os.environ['SUBJECTS_DIR']
+# subjects_dir = os.environ['SUBJECTS_DIR']
 hemi = 'lh'
 surf = 'white'
 
 # init figure
 fig = mlab.figure()
 b = Brain('fsaverage', hemi, surf, subjects_dir=subjects_dir,
-          background='white', alpha=0, figure=fig)
+          background='white', figure=fig)
 
 # co-ordinates
 [x, y, z] = b.geo[hemi].coords.T
--- a/examples/plot_custom_colors.py
+++ b/examples/plot_custom_colors.py
@@ -5,22 +5,21 @@ Plot RGBA values on brain surface
 
 """
 import os
-import nibabel
 import numpy as np
 import matplotlib.pyplot as plt
 from mayavi import mlab
 from tvtk.api import tvtk
 from tvtk.common import configure_input_data
-from surfer.utils import smoothing_matrix, mesh_edges
 from surfer import Brain
 
 print(__doc__)
 
 # 1) define helper functions
 
+
 def norm(x):
-	''' Normalise array betweeen 0-1 '''
-	return (x - np.min(x)) / (np.max(x) - np.min(x))
+    '''Normalise array betweeen 0-1'''
+    return (x - np.min(x)) / (np.max(x) - np.min(x))
 
 
 # 2) init brain and get spatial co-ordinates
--- a/examples/plot_custom_colors.py
+++ b/examples/plot_custom_colors.py
@@ -3,6 +3,11 @@
 Plot RGBA values on brain surface
 =================================
 
+In this example, each vertex on a 3D brain is plotted with a different
+RGBA value. Hue varies along the x-axis (right/left direction) and
+alpha varies along the z-axis (up/down direction). However, this can be
+easily generalised to other use cases.
+
 """
 import os
 import numpy as np
@@ -35,7 +40,7 @@ b = Brain('fsaverage', hemi, surf, subjects_dir=subjects_dir,
           background='white', figure=fig)
 
 # co-ordinates
-[x, y, z] = b.geo[hemi].coords.T
+x, y, z = b.geo[hemi].coords.T
 tris = b.geo[hemi].faces
 
 
--- a/examples/plot_custom_colours.py
+++ b/examples/plot_custom_colours.py
@@ -28,50 +28,6 @@ def norm(x, lims=[1, 99]):
     x[x > 1] = 1
     return x
 
-
-def plot_rgba_meshbrain(rgba_vals, verts, hemi='lh', surf='pial',
-                        **kwargs):
-
-    # where to load up surfaces from
-    subjects_dir = mne.datasets.sample.data_path() + '/subjects'
-
-    # make colours copy
-    colors = deepcopy(rgba_vals)
-    n_verts = colors.shape[0]
-
-    # load surface
-    rr, tris = mne.read_surface('%s/fsaverage/surf/%s.%s' % (subjects_dir,
-                                                             hemi, surf))
-    tris = tris.astype(np.uint32)
-    x, y, z = rr.T
-
-    # interpolate values from ico4 -> max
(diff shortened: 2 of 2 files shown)


agramfort:
@LauraGwilliams can you see what travis complains?

mwaskom:
I'm not sure the failures are related to this PR (looks like upstream dependency issues?) but once they're sorted out I think this is still going to be broken because pysurfer doesn't install MNE or its sample data during the CI build.

I also think including the MEG-related steps of upsampling/smoothing/etc. adds a lot of complexity to the example, which is already demonstrating something that's much more complex than any of our other scripts.

Could I please request dropping the MNE imports and simplifying the example substantially? Reading in one of our full-resolution fMRI-derived example files will skip a lot of code that seems orthogonal to adding an independent alpha channel. I'd suggest something obvious like reading in lh.sig.nii.gz and setting the alpha baed on one of the (normalized) x/y/z coordinates, or maybe using the (absolute and normalized) lh.alt_sig.nii.gz image.

larsoner:
> Could I please request dropping the MNE imports and simplifying the example substantially? Reading in one of our full-resolution fMRI-derived example files will skip a lot of code that seems orthogonal to adding an independent alpha channel. I'd suggest something obvious like reading in lh.sig.nii.gz and setting the alpha baed on one of the (normalized) x/y/z coordinates, or maybe using the (absolute and normalized) lh.alt_sig.nii.gz image.

Agreed. Really anything for demonstration purposes would work, even setting the color of the brain based on the Z coordinate value (higher is hotter!) or something.

LauraGwilliams:
OK, no problem! Is there a simple way of loading the x, y, z + triangles from a nii.gz file, such that the rest of the code can stay pretty much the same? I'm having trouble googling the answer to this.

Thank you!

larsoner:
http://nipy.org/nibabel/reference/nibabel.freesurfer.html#nibabel.freesurfer.io.read_geometry

on e.g. `SUBJECTS_DIR/subject/surf/lh.white` for white left hemi

examples/plot_custom_colours.py
```diff
@@ -0,0 +1,72 @@
...
+from surfer.utils import smoothing_matrix, mesh_edges
+from surfer import Brain
+
+print(__doc__)
+
+###############################################################################
+# define helper functions
+
+def norm(x):
+	'''
```
mwaskom: Whitespace elsewhere are spaces but these looks like tabs

examples/plot_custom_colours.py
```diff
@@ -0,0 +1,72 @@
...
+	Normalise array betweeen 0-1
+	'''
+	return (x - np.min(x)) / (np.max(x) - np.min(x))
+
+###############################################################################
+# load surface
+
+surf_fname = '/Users/lauragwilliams/Documents/programming/PySurfer/examples/example_data/lh.white'
```
mwaskom: This won't work when building the docs on travis or other people's computers, but you can build a path using `os.environ["SUBJECTS_DIR"]`

examples/plot_custom_colours.py
```diff
@@ -0,0 +1,72 @@
...
+rr, tris = nibabel.freesurfer.io.read_geometry(surf_fname)
+tris = tris.astype(np.uint32)
+x, y, z = rr.T
+
+###############################################################################
+# generate an rgba matrix, of shape n_vertices x 4
+
+# define color map
+cmap = plt.cm.rainbow
```
mwaskom: Not a good choice!

mwaskom:
The `lh.white` file is about 5Gb, which isn't huge, but we should avoid adding something that size to the code repository if possible. It should be sufficient to build a path using the `$SUBJECTS_DIR` environment variable, because anyone who is using pysurfer will have this file (and it will exist on travis because we use the basic fsaverage files to run tests).

Once the PR is finalized, I think you'll need to do a rebase so that the checked-in and then removed data file doesn't end up in master. Let us know if you need help with that.

mwaskom:
A related question for @larsoner, @agramfort : this example is revealing that several useful attributes of the `Brain` attribute (the mlab figure and the geometry) are private. This keeps the public namespace from getting too cluttered, but using them would make this (and simialr) applications even easier.  Exposing them should probably go in a different PR, but seems like a good idea?

larsoner:
e.g. `brain.geo['lh'].coords` is already public, that could be used. But sure making the figures public is reasonable. We can tell them not to modify them or things might break.

mwaskom:
> e.g. brain.geo['lh'].coords is already public, that could be used.

I'm not sure I knew about that! I'm not sure it's used in any examples ... might be good to add here.

LauraGwilliams:
I don't understand what Travis is complaining about -- am I trying to import something that's not allowed?

larsoner:
Fixed in #240 which has now been merged. If you rebase you should get meaningful feedback.

Also feel free to use `brain.geo['lh'].coords` instead of manually loading the coordinates, it will simplify things a little bit.

LauraGwilliams:
OK! Is there a way of extracting the triangles from brain.geo, too? I can only find the x, y, z...

larsoner:
`brain.geo['lh'].faces`

larsoner:
Output looks a bit weird but I think it's just a depth rendering / opacity thing:

https://118-1551431-gh.circle-artifacts.com/0/html/auto_examples/plot_custom_colours.html

If you make it opaque it might look better?

Flake errors:

https://travis-ci.org/nipy/PySurfer/jobs/386205300#L1174

larsoner:
Also let's go with `colors` instead of `colours` in the filename for consistency :)

examples/plot_custom_colors.py
```diff
@@ -26,14 +26,14 @@ def norm(x):
 # 2) init brain and get spatial co-ordinates
 
 # params
-subjects_dir = os.environ['SUBJECTS_DIR']
+# subjects_dir = os.environ['SUBJECTS_DIR']
```
larsoner: revert?

examples/plot_custom_colors.py
```diff
@@ -0,0 +1,71 @@
+"""
+=================================
+Plot RGBA values on brain surface
+=================================
+
```
agramfort: @LauraGwilliams can you add a sentence to explain what this example does?

examples/plot_custom_colors.py
```diff
@@ -0,0 +1,71 @@
...
+hemi = 'lh'
+surf = 'white'
+
+# init figure
+fig = mlab.figure()
+b = Brain('fsaverage', hemi, surf, subjects_dir=subjects_dir,
+          background='white', figure=fig)
+
+# co-ordinates
+[x, y, z] = b.geo[hemi].coords.T
```
agramfort: x, y, z = b.geo[hemi].coords.T

is enough

examples/plot_custom_colors.py
LauraGwilliams: Done, yes!

agramfort:
thx @LauraGwilliams !

LauraGwilliams:
Thank you for all your help with this!! :)

mwaskom:
Thanks again for contributing this @LauraGwilliams! I just had the opportunity to make one of these figures and it was very easy with your completed example!

LauraGwilliams:
Ah, great, I'm happy to hear @mwaskom! :D

PR #240, opened 2018-05-31
FIX: Update CIs

larsoner:
Will merge if they come back happy.


larsoner:
Lookin' good!

https://115-1551431-gh.circle-artifacts.com/0/html/auto_examples/index.html

mwaskom:
Thanks @larsoner for always keeping Travis happy

PR #244, opened 2018-06-03
FIX: AppVeyor

larsoner:


mwaskom:
image

larsoner:
AppVeyor actually works but I broke Travis by trying to simultaneously eradicate `nose` :)

But this last commit should do it!

mwaskom:
Either way I lolled pretty hard when I opened this issue and saw the commit history. Once it's running maybe you'll want to rebase the pain away :)

larsoner:
... I already had to rebase once due to merge conflict :)

But yes the "squash and merge" button is very satisfying in these situations. And the situation is actually worse than it looks, as I force-pushed a few along the way after tweaking them with `commit --amend`!

mwaskom:
BTW is nose bit rot actually breaking things or did you just want to be preemptive about jettisoning it? I'm not looking forward to having to rip it out of other projects.

larsoner:
I mostly wanted to see how difficult it would be. Turns out it was more annoying than I thought. Lots of little gotchas in there.

PR #245, opened 2018-06-27
MRG: Add auto offscreen mode

larsoner:
Adds `offscreen='auto'` as the default, which will use the `mlab.options.offscreen` value.

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -200,13 +200,17 @@ def _make_viewer(figure, n_row, n_col, title, scene_size, offscreen,
     if figure is None:
         # spawn scenes
         h, w = scene_size
-        if offscreen is True:
+        if offscreen == 'auto':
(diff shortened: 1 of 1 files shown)


larsoner:
https://161-1551431-gh.circle-artifacts.com/0/html/generated/surfer.Brain.html

agramfort:
what is it useful for?

larsoner:
For creating screenshots it's safer (sometimes on Linux you get junk images/screenshots in Mayavi with actual windows). With this I can do `mlab.options.offscreen = True` then `stc.plot(...)` without needing to add the explicit `offscreen` option to `stc.plot`.

larsoner:
(Arguably also we should have always respected the `mlab.options` value rather than defaulting to `offscreen=False`)

agramfort:
ok no objection to merge on my end

PR #247, opened 2018-07-07
Updates to install information

### LINKED ISSUE #246: Incorrect statements on install page
Just noticed that http://pysurfer.github.io/install.html claims:

> PySurfer requires Python 2.7, and it does not work on Python 3.

But that is no longer true!

Also the page starts by saying:

> Installing PySurfer is quite simple with pip:

But I think that while installing the pysurfer library is simple, in practice that claim is a bit misleading :)

mwaskom:
Closes #246


larsoner:
The CircleCI error is the same one being tackled over in https://github.com/mne-tools/mne-python/pull/5322, so once we solve it there I'll make a PR here to fix it (don't worry about it here unless you are motivated to tackle it)

doc/install.rst
```diff
@@ -5,36 +5,33 @@
...
+If you would like to save movies of time course data, it is necessary to include the optional dependency ``imageio`` with::
 
     pip install -U pysurfer[save_movie]
 
 If you'd like to install the development version, you have two options. You can
-use pip::
+install straight from github::
 
-    pip install git+git://github.com/nipy/pysurfer.git#egg=pysurfer
+    pip install git+https://github.com/nipy/pysurfer.git
```
agramfort: use:

`pip install -U https://api.github.com/repos/nipy/PySurfer/zipball/master`

so no need for git which typically a pb on windows...

doc/install.rst
```diff
@@ -44,39 +41,44 @@ To use PySurfer, you will need to have the following Python packages:
...
+Some input/output functions also make use of the Python Imaging Library (PIL_)
+and ``imageio``, although they are not mandatory.
 
 An easy option to set up this environment is the Anaconda_ distribution, which
 is free and ships with many of the required packages. If you use Anaconda,
-you'll need to install Mayavi separately. This can be done using the ``conda``
-command::
+you'll need to install Mayavi separately. This should be done using ``conda``::
 
     conda install mayavi
```
agramfort: I doubt this actually works beyond 2.7. @larsoner we should use the install procedure as explained for MNE.

christianbrodbeck:
Actually for any platform except macOS/Python 3 the whole installation should be possible through conda (`conda install pysurfer`).

mwaskom:
I’m traveling the next few days (and not an expert on installing) so feel free to push changes straight to the branch.

doc/install.rst
```diff
@@ -20,7 +20,7 @@ If you would like to save movies of time course data, it is necessary to include
 If you'd like to install the development version, you have two options. You can
 install straight from github::
 
-    pip install git+https://github.com/nipy/pysurfer.git
+    pip install -U https://api.github.com/repos/nipy/PySurfer/zipball/master
```
mwaskom: If we're gonna put core numeric packages in ``install_requires`` we might want to add ``--no-deps`` here so people aren't upgrading lots of things they might not want to be.

larsoner:
For now I removed the Anaconda instructions, and tell users to look it up. This seems preferable to giving the MNE instructions, because a) those are currently broken and b) I don't want to have to keep them up to date all over the place. I'd rather just wait until Anaconda actually has a `mayavi` + 3.6 solution that works for all platforms.

mwaskom:
LGTM @larsoner

larsoner:
Okay will merge once the CIs are happy. If others have improvements we can always make incremental changes

mwaskom:
Thanks @larsoner

PR #248, opened 2018-08-01
MRG: Add TimeViewer to docs

larsoner:
@OlehKSS noticed that `TimeViewer` was not in the API docs. This adds it.


larsoner:
https://187-1551431-gh.circle-artifacts.com/0/html/python_reference.html

PR #251, opened 2018-08-30
[MRG] ADD:  Brain.remove_foci()

christianbrodbeck:
Seemed like it was missing and simple enough to add.

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -1454,6 +1454,33 @@ class Brain(object):
         if all(len(brain.data) == 0 for brain in self.brains):
             self.n_times = self._times = None
 
+    def remove_foci(self, foci=None):
+        """Remove foci added with ``Brain.add_foci()``.
+
+        Parameters
+        ----------
+        foci : str | Sequence of str
+            Names of the foci to remove (default all).
+
+        Notes
+        -----
+        Only foci added with a unique names can be removed.
+        """
+        if foci is None:
+            keys = tuple(self.foci_dict)
+        else:
+            if isinstance(foci, str):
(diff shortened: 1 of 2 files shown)


christianbrodbeck:
Yes, `remove_data()` and `remove_labels()`, glad you asked :)

christianbrodbeck:
We could add a new `Brain` attribute to keep track of unnamed foci too, but it would add some more overhead...

surfer/viz.py
```diff
@@ -1454,6 +1454,34 @@ def remove_data(self, hemi=None):
         if all(len(brain.data) == 0 for brain in self.brains):
             self.n_times = self._times = None
 
+    def remove_foci(self, foci=None):
```
agramfort: please call it name to be consistent with add_foci API.

surfer/viz.py
```diff
@@ -1454,6 +1454,34 @@ def remove_data(self, hemi=None):
...
+        """Remove foci added with ``Brain.add_foci()``.
+
+        Parameters
+        ----------
+        foci : str | Sequence of str
```
agramfort: str | list of str

agramfort: or better:

str | list of str | None

surfer/viz.py
```diff
@@ -1454,6 +1454,34 @@ def remove_data(self, hemi=None):
...
+            Names of the foci to remove (default all).
```
agramfort: If None, remove all focis.

larsoner:
Thanks @christianbrodbeck

PR #252, opened 2018-09-07
MRG: Deprecated args

larsoner:
Fix use of deprecated function.


larsoner:
Replaced deprecated use of `inspect.getargspec` and also fixed the CIs. Ready for review/merge from my end.

PR #253, opened 2018-09-20
FIX: annotations and bgcolor

### LINKED ISSUE #250: brain.add_annotation does not display annotation
Hi,

I am trying to display Freesurfer annotation using brain.add_annotation('aparc'). However, pysurfer does not display the annotation. I'm using Mac OS and qt backend. I appreciate any input on this. 

thanks
Ehsan.

larsoner:
Closes #250.

Works around a `nibabel` bug and actually should slightly improve the `borders` functionality by setting the transparent color better (to the brain background with alpha 0 instead of black with alpha 0, which helps with transitions).

@fliem can you see if this fixes your problem?

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -1267,13 +1267,11 @@ class Brain(object):
             self._to_borders(labels, hemi, borders)
 
             # Handle null labels properly
-            # (tksurfer doesn't use the alpha channel, so sometimes this
-            # is set weirdly. For our purposes, it should always be 0.
-            # Unless this sometimes causes problems?
             cmap[:, 3] = 255
             bgcolor = self._brain_color
             bgcolor[-1] = 0
-            cmap[np.where(cmap[:, 4] <= 0), :4] = bgcolor
+            cmap[cmap[:, 4] < 0, 4] += 2 ** 24  # wrap to positive
+            cmap[cmap[:, 4] <= 0, :4] = bgcolor
             if np.any(labels == 0) and not np.any(cmap[:, -1] <= 0):
                 cmap = np.vstack((cmap, np.concatenate([bgcolor, [0]])))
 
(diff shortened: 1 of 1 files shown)


larsoner:
Renders okay

https://222-1551431-gh.circle-artifacts.com/0/html/auto_examples/plot_parcellation.html

fliem:
Thanks @larsoner. Works with nibabel 2.3.0

agramfort:
works like a charm for me !

larsoner:
Updated to deal with future considerations (negative IDs) mentioned in https://github.com/nipy/nibabel/issues/649. Still renders fine:

https://226-1551431-gh.circle-artifacts.com/0/html/auto_examples/plot_parcellation.html

Ready for review/merge from my end.

Then maybe one of these days we should cut a new release!

surfer/viz.py
```diff
@@ -1885,6 +1886,21 @@ def set_surf(self, surf):
...
+            bgcolor = np.mean(
+                self._brain_list[0]['brain']._geo_surf.module_manager
+                .scalar_lut_manager.lut.table.to_array(), axis=0)
+        else:
+            bgcolor = geo_actor.property.color
+            if len(bgcolor) == 3:
+                bgcolor = bgcolor + (1,)
+            bgcolor = 255 * np.array(bgcolor)
+        bgcolor[-1] *= geo_actor.property.opacity
+        return bgcolor
```
agramfort: I don't like to see properties that do actual work although I agree here it's unlikely to be called in a loop.

larsoner: This seems roughly consistent with what is already done in PySurfer -- it's mostly about triaging values rather than doing any big calculations

agramfort:
yes feel free to consider this as a nitpick

+1 for MRG

mwaskom:
I read the diff quickly so I might be missing something, but if you have an annotation plotted on top of a statistical map, isn't this going to make the undefined vertices in the annotation show up with the base geometry colors, even if the statistical map is defined there?

larsoner:
No the alpha value is zero. It just makes the tiny transition region (where colors are interpolated by OpenGL) between the opaque border and whatever is under it gray instead of black

mwaskom:
Ah, got it.

PR #259, opened 2018-10-20
Mapping values to foci.

cristidonos:
I made the following changes to plot SEEG electrodes and task-related activations:
- Brain.add_foci can now map values using foci color and size.
- added Brain.add_text3d for labeling foci in 3d

foci
labels

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -1570,7 +1570,7 @@ class Brain(object):
 
     def add_foci(self, coords, coords_as_verts=False, map_surface=None,
                  scale_factor=1, color="white", alpha=1, name=None,
-                 hemi=None):
+                 hemi=None, allow_data=False, data=None):
         """Add spherical foci, possibly mapping to displayed surf.
 
         The foci spheres can be displayed at the coordinates given, or
@@ -1588,8 +1588,8 @@ class Brain(object):
             whether the coords parameter should be interpreted as vertex ids
         map_surface : Freesurfer surf or None
             surface to map coordinates through, or None to use raw coords
-        scale_factor : float
-            Controls the size of the foci spheres (relative to 1cm).
+        scale_factor : int
+            controls the size of the foci spheres
         color : matplotlib color code
             HTML name, RBG tuple, or hex code
         alpha : float in [0, 1]
@@ -1600,6 +1600,14 @@ class Brain(object):
             If None, it is assumed to belong to the hemipshere being
             shown. If two hemispheres are being shown, an error will
             be thrown.
+        allow_data : bool | False
+            If True, will plot foci using quiver3d instead of points3d to allow
+            mapping data to spheres.
+        data : numpy array
+            1D array the same size as the number of foci. Spheres will be
+            colorcoded acording to data values, whereas spheres' sizes will be
+            coded using the absolute values of data.
+
         """
         from matplotlib.colors import colorConverter
         hemi = self._check_hemi(hemi)
@@ -1614,8 +1622,7 @@ class Brain(object):
             foci_coords = np.atleast_2d(coords)
         else:
             foci_surf = Surface(self.subject_id, hemi, map_surface,
-                                subjects_dir=self.subjects_dir,
-                                units=self._units)
+                                subjects_dir=self.subjects_dir)
             foci_surf.load_geometry()
             foci_vtxs = utils.find_closest_vertices(foci_surf.coords, coords)
             foci_coords = self.geo[hemi].coords[foci_vtxs]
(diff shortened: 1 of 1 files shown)


surfer/viz.py
```diff
@@ -1600,6 +1600,14 @@ def add_foci(self, coords, coords_as_verts=False, map_surface=None,
             If None, it is assumed to belong to the hemipshere being
             shown. If two hemispheres are being shown, an error will
             be thrown.
+        allow_data : bool | False
+            If True, will plot foci using quiver3d instead of points3d to allow
+            mapping data to spheres.
+        data : numpy array
+            1D array the same size as the number of foci. Spheres will be
+            colorcoded acording to data values, whereas spheres' sizes will be
+            coded using the absolute values of data.
```
agramfort: API seems quite complicated. Can you achieve what you want using only one new parameter? and without adding a new method?

also you miss tests and some new example for documentation.

surfer/viz.py
```diff
@@ -1588,8 +1588,8 @@ def add_foci(self, coords, coords_as_verts=False, map_surface=None,
             whether the coords parameter should be interpreted as vertex ids
         map_surface : Freesurfer surf or None
             surface to map coordinates through, or None to use raw coords
-        scale_factor : float
-            Controls the size of the foci spheres (relative to 1cm).
+        scale_factor : int
```
mwaskom: Why the change to int here? Scale factors can be floats, as seen in the example: http://pysurfer.github.io/auto_examples/plot_foci.html Do your changes break that, or is this just an errant attempt to fix the documentation?

surfer/viz.py
mwaskom: I agree with @agramfort. Can `allow_data` as an external parameter not be replaced with an internal check for `data is not None`?

surfer/viz.py
```diff
@@ -1737,6 +1748,45 @@ def add_text(self, x, y, text, name, color=None, opacity=1.0,
         if justification is not None:
             text.property.justification = justification
 
+    def add_text3d(self, x, y, z, text, name, color=None, opacity=1.0,
+                 row=-1, col=-1, font_size=None, justification=None):
```
mwaskom: Do we use this `row`, `col` parameterization anywhere else in pysurfer? Ideally if you're going to add a new method it should match the existing API as closely as possible.

surfer/viz.py
```diff
@@ -1737,6 +1748,45 @@ def add_text(self, x, y, text, name, color=None, opacity=1.0,
...
+        """
+        if name in self.texts_dict:
+            self.texts_dict[name]['text'].remove()
+        text = self.brain_matrix[row, col].add_text3d(x, y, z, text,
+                                                    name, color, opacity)
+        self.texts_dict[name] = dict(row=row, col=col, text=text)
+        if font_size is not None:
+            text.property.font_size = font_size
+            text.actor.text_scale_mode = 'viewport'
+        if justification is not None:
```
mwaskom: This parameter doesn't appear in the docstring. Also despite it being the property name in mayavi, "justification" is a weird term and "alignment" would be more correct.

surfer/viz.py
mwaskom: I also am not sure that I like forcing the color *and* size of the spheres by default. There are lots of cases when you might want colors to reflect a scalar but a uniform size. Or (probably less commonly) vice versa. Also I don't see any way to control the size range, colormap, etc. through this API, which would be important.

cristidonos: Thanks for pointing out the issues, i'll try to fix them. This is my first attempt at a pull request, so bear with me, especially with the tests.

cristidonos:
Hi,
Made some changes, not there's only one add_foci function that uses quiver3d instead of points3d to draw the spheres. Implemented internal checks for data and made color-coded spheres optional. 

Regarding text functions: i only copied add_text function from above and changed it to uses the text3d function in mayavi. All 4 arguments (row, col, justification and font_size) were already there. 

Couldn't figure out failed tests myself, please help me fix those.

danielegrattarola:
Hi, was this feature ever implemented in the library?
I need to plot some text on a brain surface but I'm not sure how to achieve this in pysurfer. 

Thanks

mwaskom:
@danielegrattarola This has not been merged. There are unresolved questions about how best to integrate into the existing PySurfer API, and the tests are not passing. But looking at the code should give you a pretty good idea of how to add text onto a PySurfer visualization. It will take you writing more code than if it were just a method on the `Brain` object, but it is definitely possible.

danielegrattarola:
@mwaskom Thanks, I've managed to hack a solution together using `mlab.text3d` for now. 
Cheers

PR #261, opened 2018-10-25
MRG: Fix scale_data_colormap and add FXAA

### LINKED ISSUE #260: Displaying multiples sources estimates mixes up different colormaps
Hi, 

I'm trying to display source estimates of different subjects on the same brain surface using different colormaps. However, some sources appear to switch colors during the process. `brain.add_data` calls do not seem to be independent. In this snippet, I simulate sources on both hemispheres for 2 subjects. I plot sources independently for each subject and then jointly on the same brain surface. 2 red sources  (up) get colored in blue when displayed on the same brain. Sources are taken to be non-overlapping across subjects. 

```python
import numpy as np  # noqa
import mne
import os

from mne.datasets.sample import data_path
from mayavi import mlab  # noqa
from surfer import Brain  # noqa

data_path = data_path()
subjects_dir = data_path + "/subjects/"
os.environ['SUBJECTS_DIR'] = subjects_dir

subject = "fsaverage"
fname_fs = subjects_dir + "fsaverage/bem/fsaverage-ico-5-src.fif"
src_fs = mne.read_source_spaces(fname_fs)
nv = 1000  # keep only n_v vertices per hemi
vertices = [src_fs[0]["vertno"][:nv], src_fs[1]["vertno"][:nv]]

n_subjects = 2
sources_l = np.zeros((nv, n_subjects))
sources_r = np.zeros((nv, n_subjects))

# subject 1
sources_l[:10, 0] = 5
sources_r[:10, 0] = 3

# subject 2
sources_l[100:110, 1] = 2
sources_r[100:110, 1] = 5

sources = [sources_l, sources_r]
colormaps = ["Reds", "Blues"]


def plot_sources(sources, sub_id=[0, 1], order=1):
    fmax = 5
    brain = Brain(subject, hemi="both", surf="inflated", views="dorsal")
    hemis = ["lh", "rh"]
    for sources_h, v, h in zip(sources[::order], vertices[::order],
                               hemis[::order]):
        for data, colormap in zip(sources_h.T[sub_id],
                                  np.array(colormaps)[sub_id]):
                brain.add_data(data, colormap=colormap, vertices=v,
                               verbose=False, colorbar=False,
                               smoothing_steps=10,
                               time_label=None, hemi=h, alpha=0.8,
                               min=0., mid=fmax / 5, max=fmax,
                               transparent=True)
    return brain
```

screen shot 2018-10-25 at 12 25 40

Done with pysurfer 0.10.dev0 and mayavi 4.5.1.dev0.

larsoner:
Closes #260.

@hichamjanati can you see if this solves your problem?

I also added a `use_fxaa = True` to improve rendering quality slightly while I was in there, and fixed where we should have enabled backface culling for the quiver arrows.

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -229,6 +229,9 @@ def _make_viewer(figure, n_row, n_col, title, scene_size, offscreen,
                     for f in figure:
                         f.scene.interactor.interactor_style = \
                             tvtk.InteractorStyleTerrain()
+            for figure in figures:
+                for f in figure:
+                    f.scene.renderer.use_fxaa = True
     else:
         if isinstance(figure, int):  # use figure with specified id
             figure = [mlab.figure(figure, size=scene_size)]
@@ -387,6 +390,7 @@ class Brain(object):
     texts : dict
         The text objects.
     """
+
     def __init__(self, subject_id, hemi, surf, title=None,
                  cortex="classic", alpha=1.0, size=800, background="black",
                  foreground=None, figure=None, subjects_dir=None,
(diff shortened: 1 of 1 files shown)


hichamjanati:
Thanks @larsoner, it seems to be fixed now.

larsoner:
Because the `testing` mayavi backend is incomplete, should be a quick fix

PR #265, opened 2019-03-22
MRG: width/height in offscreen mode

larsoner:
Fixes a bug from #245 that showed up when `w != h` as values ended up swapped.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,53 +1,66 @@
 version: 2
 jobs:
-    build:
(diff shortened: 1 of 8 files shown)


larsoner:
Ready for review/merge from my end. This one is a pretty trivial fix aside from the necessary CI updating. I'll merge tomorrow probably assuming it comes back green unless someone wants more time to look.

agramfort:
thx @larsoner

Issue #134, opened 2015-06-13
Support for use of alternate atlases (e.g. Human Connectome Project atlas)


bcipolli:
Tools / datasets from the Van Essen lab, such as the Human Connectome Project (HCP), Connectome workbench, and SumsDB, make surface data readily available without using a freesurfer installation. These include alternate atlases with different vertex counts.

Currently PySurfer assumes a lot--it assumes a freesurfer-like directory structure for data, it loads curvature information from strongly assumed file locations (base directory path is flexible, but the subdirectory and filename are assumed).

To generalize things and make the tools more useful for the new HCP workflows, I suggest generalizing all paths to be taken as optional arguments. The defaults can be built from the assumed freesurfer locations, but I would like to specify my own atlas (in addition to my own overlays).

I'm hacking a prototype together for this for OHBM (as I've got HCP surface data that needs to be plotted on HCP atlases), but wanted to get interests / comments on the idea.

Thanks!

mwaskom:
> To generalize things and make the tools more useful for the new HCP workflows, I suggest generalizing all paths to be taken as optional arguments. The defaults can be built from the assumed freesurfer locations, but I would like to specify my own atlas (in addition to my own overlays).

IMO this seems radically easier said than done. PySurfer is very closely tied to Freesurfer, not just in the location of the data, but also in what things are called ("annotations", "labels", "morphometry"), etc. Taking what we have and extending it to work with an entirely different sphere of data, with its own conventions and quirks, seems destined to be a massive hack.

I agree that it would be great to have a nice Pythonic interface to visualize HCP-style surface data. I'm not sure that building flexibility into PySurfer is the right way to do it, though.

bcipolli:
The HCP data use freesurfer and file formats readable by nibabel. It's easy to convert file formats before calling into PySurfer (or using the power of nibabel to read files and make PySurfer agnostic to format); the problem is the directory structure and file naming assumptions.

I believe there's a relatively general way to do this (passing file paths to needed files, rather than passing a `subjects_dir` and assuming you know how to construct the paths). All the file formats can be assumed still, in the way I'm thinking.

Does that sound more reasonable?

dengemann:
Note that we have a related discussion elsewhere:
https://github.com/mne-tools/mne-python/issues/2116

cc @mwaskom

dengemann:
PySurfer would be part of the solution somehow or we will eventually find another viz backend.

bcipolli:
[...]

Issue #106, opened 2014-05-08
backends and multithreading


aestrivex:
The current pysurfer code to use traitsui involves strange hacks to make the GUI busy for a short time while X renders the mayavi window but also without causing the GUI to enter an event loop and block. This section of code is very difficult to understand and maintain. A much more natural solution would be to use multithreading and explicit management of locking constructs to ensure that all of the events happen at the correct time.

The problem with this code is mainly with traitsui, which is an interactive GUI but which pysurfer is trying to use as a noninteractive GUI. Traitsui doesn't expose an API to usefully allow for precise timing of the GUI to be controlled in a multithreaded environment. In order to get the correct behavior without generating potential race conditions (which I think is what is happening now, only it is a race condition that is extremely unlikely to ever lose the race), I don't think embedding mayavi in traitsui will work.

I don't know any visualization libraries other than mayavi but I have been working on some simple tests to operate on VTK windows can be managed and display at the correct times in GUI code that does not involve traitsui -- I have proof of concept tests that suggest this should work with wxpython and tk (but it did not work at all with pyqt). As wxpython is not included in anaconda by default (and nothing about the GUI layout in pysurfer is particularly complex), tk is probably the better choice.

I'm of the opinion that to fix the current problems with traitsui, pysurfer should move in this direction (and I anticipate doing the majority of the work). This won't per se help eliminate dependencies on python 2.7 more quickly, but it will provide something more stable and maintainable so that pysurfer might later move to multiple backends as has been discussed before.

wmvanvliet:
@Eric89GXL any progress on the move to vispy? :)

Eric89GXL:
Heh, not much so far

mwaskom:
Meanwhile, looks like some progress on getting mayavi running on python 3: https://github.com/enthought/mayavi/pull/250

wmvanvliet:
that's promising!

choldgraf:
Hey folks - diving in this conversation with another potential option here. I've been working on the `plotly` trisurf libraries for the last week, and got the plotting time for an `fsaverage` hemisphere down to about 10 seconds. It's still a bit slow IMO, but probably there are lots of improvements to be made as well.

Plotly uses WebGL as a backend, so the plots are rendered in the browser but they're pretty smooth and much more stable than Mayavi. E.g., here's a quick package/demo I put together as a proof of concept:

https://github.com/choldgraf/ecogtools/blob/master/examples/3d_plots.ipynb

[...]

PR #268, opened 2019-05-21
Integrate better with Jupyter notebook

wmvanvliet:
Mayavi can render things inside a jupyter notebook as either PNG or X3D.
This functionality can be enabled with `mlab.init_notebook()`.

This PR adds an `_ipython_display_` hook to the `Brain` class that
renders the brain accordgin to the Mayavi notebook integration settings.

This means interactive PySurfer brains in Binder are now possible!
https://mybinder.org/v2/gh/wmvanvliet/SNL_workshop_2019/master?filepath=pysurfer_demo.ipynb

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -78,14 +78,15 @@ This will allow you to have an open PySurfer window while still being able to
 execute code in the console/notebook.
 
 It is also possible to embed the PySurfer visualization into a Jupyter notebook.
-This is achieved by leveraging Mayavi's notebook integration::
+This is achieved by leveraging `Mayavi's notebook integration
+<https://docs.enthought.com/mayavi/mayavi/tips.html#using-mayavi-in-jupyter-notebooks>`_::
 
     from mayavi import mlab
     mlab.init_notebook(backend='png')
 
 The ``backend`` parameter can either be ``'png'`` to render the visualization
 as a static PNG image, or ``'x3d'`` to render it using
-`X3D <https://www.x3dom.org>` (still experimental).
+`X3D <https://www.x3dom.org>`_ (still experimental).
 
 If you are having trouble getting started using PySurfer, please describe the problem on the `nipy mailing list`_.
 
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -77,6 +77,16 @@ notebook), you have to activate the correct GUI backend, which is probably qt::
 This will allow you to have an open PySurfer window while still being able to
 execute code in the console/notebook.
 
+It is also possible to embed the PySurfer visualization into a Jupyter notebook.
+This is achieved by leveraging Mayavi's notebook integration::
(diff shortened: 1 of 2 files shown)


larsoner:
> This means interactive PySurfer brains in Binder are now possible!

This would be through the `X3D` backend? It's mature enough that you can for example rotate the brain with the mouse?

larsoner:
Oh that's a `mybinder` link, trying now...

larsoner:
It interacts but has some bugs:

Screenshot from 2019-05-21 10-02-26

We can probably try to fix this at the `mayavi` end, this looks like a hopefully easy bytes vs str issue.

In any case, +1 for merge from me!

banesullivan:
You could also achieve this with PyVista. Cross-post from https://github.com/pyvista/pyvista/issues/146#issuecomment-494406647

Check out all the PyVista examples on http://playgroiund.pyvista.org

Or this one: https://mybinder.org/v2/gh/OpenGeoVis/omfvista/master?filepath=Example.ipynb

omfvtk

This is enabled by default if [`panel`](https://panel.pyviz.org) is installed

mwaskom:
It would be good to have this mentioned in the documentation. Maybe at the bottom of [this page](http://pysurfer.github.io/install.html)?

larsoner:
BTW @wmvanvliet I don't hit the bytes/str bug locally, so if you update your `mayavi` on MyBinder to latest master it will probably go away

jasmainak:
neat @wmvanvliet !

wmvanvliet:
The `str` vs `bytes` bug is actually an empty string bug. Sometimes VTK is too slow to render the brain and the X3D string becomes empty. I also hit it locally sometimes. It's a mayavi/VTK bug and generally the X3D is kind of unstable. However, it's better than nothing when in a pinch :)

wmvanvliet:
> You could also achieve this with PyVista
Time to rewrite pysurfer in pyvista? ;)

larsoner:
> Time to rewrite pysurfer in pyvista? ;)

I know you were joking, but just in case you're curious -- we've had long discussions about this in #106 and #134. IIRC the consensus was basically that we should start some new neuro-3D-viz package that is more general, and less bound to Freesurfer conventions (but still provides what we need for working with FS).

To make some progress in MNE, @GuillaumeFavelier has made some progress implementing Mayavi alternatives (PyVista, ipyvolume, VisPy) for our basic 3D functionality such as `plot_alignment` working. One of the next big targets will be source estimate plotting (currently we use PySurfer for this). Once things actually work to some extent, it will probably make sense to spin this off into its own package.

larsoner:
In the meantime this is great, thanks @wmvanvliet

mwaskom:
Thanks @wmvanvliet!

Issue #214, opened 2017-10-03
ENH: Flatmaps


larsoner:
I'm interested in doing flatmaps with PySurfer. Has anyone tried this?

A very naive approach did not work:
```
>>> brain = surfer.Brain('fsaverage', 'lh', 'cortex.patch.flat')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/larsoner/custombuilds/PySurfer/surfer/viz.py", line 451, in __init__
    geo.load_geometry()
  File "/home/larsoner/custombuilds/PySurfer/surfer/utils.py", line 126, in load_geometry
    coords, faces = nib.freesurfer.read_geometry(surf_path)
  File "/home/larsoner/miniconda3/envs/mne/lib/python3.6/site-packages/nibabel/freesurfer/io.py", line 128, in read_geometry
    quads = quads.reshape(nquad, 4)
ValueError: cannot reshape array of size 0 into shape (4665600,4)
```

mwaskom:
Patch files are a completely different format.

I have a notebook where I got part of the way: https://gist.github.com/mwaskom/fab1970208a5c458869b3d8831fb7788

In general I'd think plotting on flat maps would be better handled by matplotlib than mavavi.

larsoner:
> Patch files are a completely different format.

I'm guessing there is no simple mapping to vertices, then :(

> In general I'd think plotting on flat maps would be better handled by matplotlib than mavavi.

Yeah that's true. And then Mayavi wouldn't be a blocker. PySurfer does a ton of stuff (`add_data`, time steps, videos, etc.) that would be nice to have in both cases. We could consider using a `matplotlib` backend when using the flat maps, but that would be quite a big degree of added difficulty :)

I probably won't dig into it now, then, since it's not so simple.

PR #269, opened 2019-06-24
Adding flatmap rendering functionallity

Tal-Golan:
Addressing [issue 214](https://github.com/nipy/PySurfer/issues/214#issue-262540068), I added flat-map rendering functionality. This was done by creating a Patch, a child object of Surface, that deal with storing the relevant meta-data (which patch vertices correspond to which surface vertices). I worked over all of the examples and modified the relevant functions so they work with flatmaps as seamlessly as possible.


examples/plot_custom_colors.py
```diff
@@ -74,3 +74,28 @@ def norm(x):
...
+    x, y, z, tris, figure=fig)
+mesh.data.point_data.scalars.number_of_components = 4  # r, g, b, a
+mesh.data.point_data.scalars = (rgba_vals * 255).astype('ubyte')
+
+# tvtk for vis
+mapper = tvtk.PolyDataMapper()
+configure_input_data(mapper, mesh.data)
+actor = tvtk.Actor()
+actor.mapper = mapper
+fig.scene.add_actor(actor)
```
larsoner: I would not modify every example, just one or two. Maybe adding the `plot_flatmap.py` is already enough.

examples/plot_flatmap.py
```diff
@@ -0,0 +1,26 @@
...
+brain.add_label(label='V1_exvivo',hemi='lh')
+brain.add_label(label='V1_exvivo',hemi='rh')
+
+overlay_file = "example_data/lh.sig.nii.gz"
+brain.add_overlay(overlay_file,hemi='lh')
+
+cam = fig.scene.camera
+cam.zoom(1.85)
+
+mlab.savefig('test.png',figure=fig,magnification=5) # save a high-res figure
```
larsoner: No need for this `savefig` command in the example

examples/plot_fmri_activation_volume.py
```diff
@@ -76,5 +76,5 @@
 mask = ~mask
 brain.add_data(mask, min=0, max=10, thresh=.5,
                colormap="bone", alpha=.6, colorbar=False)
-
-brain.show_view("medial")
+if not hasattr(brain,'patch_mode'):
+    brain.show_view("medial")
```
larsoner: I would just remove this

examples/plot_freesurfer_normalization.py
```diff
@@ -34,4 +34,5 @@
                               line_width=3, hemi=hemi)
     brain.contour_list[-1]["colorbar"].visible = False
 
-brain.show_view("dorsal")
+if not brain.patch_mode:
+    brain.show_view("dorsal")
```
larsoner: Instead of forcing people to write conditionals, in `brain.show_view` I would check to see if in `patch_mode` and if so, discard the azimuth/elevation/roll arguments (only pay attention to `distance`), and say that this is the behavior in the `show_view` docstring. If people *really* want to configure the view they can use `mlab` to do it.

examples/plot_parcellation.py
```diff
@@ -14,7 +14,10 @@
 subject_id = 'fsaverage'
 hemi = 'both'
 surf = 'inflated'
-view = 'frontal'
+if 'patch' in surf.lower().split('.'):
+    view=None
+else:
+    view = 'frontal'
```
larsoner: None of these conditionals are needed with the `show_view` modification suggested above.

surfer/utils.py
```diff
@@ -175,6 +175,201 @@ def apply_xfm(self, mtx):
...
+        orig_coords, orig_faces = nib.freesurfer.read_geometry(surf_path)
+        n_orig_vertices=orig_coords.shape[0]
+        assert np.max(patch['vno']) < n_orig_vertices, 'mismatching vertices in patch and orig surface'
+
+        # re-define faces to use the indecis of the selected vertices
+        patch_vertices_in_original_surf_indexing=patch['vno']
+
+        # reverse the lookup table:
+        original_vertices_in_patch_indexing=np.zeros((n_orig_vertices,)); original_vertices_in_patch_indexing[:]=np.nan
+        original_vertices_in_patch_indexing[patch_vertices_in_original_surf_indexing]=np.arange(len(patch_vertices_in_original_surf_indexing))
```
larsoner: There are many PEP8 violations in this and other files, can you fix them?

surfer/utils.py
```diff
@@ -175,6 +175,201 @@ def apply_xfm(self, mtx):
...
+        vertices=vertices[vertices_in_patch]
+        vertices=np.array(vertices,original_dtype)
+        if len(args)==0:
+            return vertices
+        else:
+            cut_v=[]
+            for v in args:
+                cut_v.append(np.asarray(v)[vertices_in_patch])
+            return (vertices,)+tuple(cut_v)
+def read_patch_file(fname):
```
larsoner: Is this in `nibabel`? If not, can you make a PR there, too?

surfer/viz.py
```diff
@@ -361,7 +363,9 @@ class Brain(object):
         views to use
     offset : bool
         If True, aligs origin with medial wall. Useful for viewing inflated
-        surface where hemispheres typically overlap (Default: True)
+        surface where hemispheres typically overlap. For flat patches,
+        this aligns the two patches along their posterior
+        edges. (Default: True)
```
larsoner: You should also update the `surface` argument above to say that it's supported, and add a:
```

.. versionchanged:: 0.10
   Support for flat maps.
```

larsoner:
Now that #270 is in, if you rebase CircleCI might render properly.

PR #270, opened 2019-06-25
MAINT: Update fsaverage

larsoner:
In MNE we recently introduced a `fsaverage` dataset fetcher. Might as well use it here. It should help with #269.


larsoner:
Bumps minimum Travis run to 3.5 since MNE needs it. Probably not a bad time to move to 3.5+ anyway since most Python packages are doing that nowadays.

larsoner:
Finally ready for review/merge. Maybe @agramfort ?

https://267-1551431-gh.circle-artifacts.com/0/html/auto_examples/index.html

larsoner:
Okay let's get this in so that #269 can benefit

PR #272, opened 2019-07-30
MAINT: Fix ignores for CIs

larsoner:


larsoner:
AppVeyor backlog is insane, this fixes Travis so merging (it should also fix AppVeyor but if it doesn't I'll just open another PR)

PR #273, opened 2019-08-12
ENH: Allow painting all annotation vertices in a single color

mwaskom:
A small enhancement that makes it easy to show the borders between annotation labels in a single color. This makes less distracting figures in cases where you want to overlay an annotation on a statistical map and either know what each parcel represents or don't precisely care.

e.g.

```python
from surfer import Brain
b = Brain("fsaverage", "lh", "inflated")
b.add_annotation("aparc", color="dodgerblue")
```
annot_edges

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -1196,7 +1196,7 @@ class Brain(object):
         self._toggle_render(True, views)
 
     def add_annotation(self, annot, borders=True, alpha=1, hemi=None,
-                       remove_existing=True):
+                       remove_existing=True, color=None):
         """Add an annotation file.
 
         Parameters
@@ -1220,6 +1220,10 @@ class Brain(object):
             for both hemispheres.
         remove_existing : bool
             If True (default), remove old annotations.
+        color : matplotlib-style color code
+            If used, show all annotations in the same (specified) color.
+            Probably useful only when showing annotation borders.
+
         """
         hemis = self._check_hemis(hemi)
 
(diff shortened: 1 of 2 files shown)


surfer/viz.py
```diff
@@ -1292,6 +1296,12 @@ def add_annotation(self, annot, borders=True, alpha=1, hemi=None,
             alpha_vec = cmap[:, 3]
             alpha_vec[alpha_vec > 0] = alpha * 255
 
+            # Override the cmap when a single color is used
+            if color is not None:
+                from matplotlib.colors import colorConverter
+                rgb = np.round(np.multiply(colorConverter.to_rgb(color), 255))
+                cmap[:, :3] = rgb.astype(cmap.dtype)
```
larsoner: Should it be `:4` and `to_rgba`?

mwaskom: There's already a parameter in `add_annotation` that controls the opacity of the overlay. I thought it made the most sense to continue honoring that argument and to ignore any alpha channel in the color spec itself.

In other words, like this

```
b.add_annotation("aparc", color="forestgreen", alpha=.5)
```
annot_edges

I guess one could check to see if the color has an alpha channel and error or warn if it does not match the alpha specified in the function call. But seems like overkill?

larsoner:
Agreed, thanks @mwaskom 

Travis error is https://github.com/ContinuumIO/anaconda-issues/issues/11195 so let's ignore it

mwaskom:
Thanks for the quick merge @larsoner

PR #274, opened 2019-09-06
FIX:  don't change view when adding foci

christianbrodbeck:


larsoner:
Makes sense, thanks @christianbrodbeck

PR #275, opened 2019-09-24
ENH: Add option to smooth to nearest

larsoner:
Allows doing `smoothing_steps='nearest'` so you can see the patches of cortex like (not the most relevant example but gives the idea):

Screenshot from 2019-09-24 12-26-44

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .#*
 *.swp
 *.orig
+*.mov
 build
 
 dist/
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,24 @@
-PySurfer Changes
-================
+Changelog
+=========
+
+.. currentmodule:: surfer
+
+Development version (0.10.dev0)
+-------------------------------
+
+- Added an option to smooth to nearest vertex in :meth:`Brain.add_data` using
+  ``smoothing_steps='nearest'``
+- Added options for using offscreen mode
+- Improved integration with Jupyter notebook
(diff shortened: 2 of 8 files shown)


mwaskom:
Maybe a better example use-case would be doing some kind of discrete labeling in low-res space and then propagating the labels to the high-res surface?

larsoner:
>  some kind of discrete labeling in low-res space and then propagating the labels to the high-res surface?

That's basically what a MNE inverse solution is, a low(er)-res subsampling of the high res surface and then when we plot we upsample.

Ultimately this might not even need an example, so I can revert it if need be. I'd rather not write a separate new one just for this.

larsoner:
Got the little indexing demons worked out, this is what the MEG one looks like with `'nearest'`

https://281-1551431-gh.circle-artifacts.com/0/html/auto_examples/plot_meg_inverse_solution.html

mwaskom:
That's right — just the [changes file](https://github.com/nipy/PySurfer/blob/master/CHANGES) that is usually the basis for a release announcement email.

larsoner:
That file only goes up to 0.8, and we've released 0.9 already. I guess we're a bit behind there. I can add a 0.9 and dev section here sometime today or tomorrow

larsoner:
Feast your eyes on the beautiful changelog:

https://282-1551431-gh.circle-artifacts.com/0/html/changes.html

agramfort:
I have rarely been so pleased...

>

PR #276, opened 2019-10-03
ENH: Add kwargs in Brain.add_* and pass it to mayavi modules

DingkunLiu:
Add kwargs in Brain.add_* and pass it to mayavi modules, so that we could directly manipulate mayavi display settings, like point resolution and overlay opacity.


surfer/viz.py
```diff
@@ -932,6 +932,7 @@ def add_overlay(self, source, min=2, max="robust_max", sign="abs",
             If None, it is assumed to belong to the hemipshere being
             shown. If two hemispheres are being shown, an error will
             be thrown.
+        kwargs: other mayavi surface arguments.
```
larsoner: This should follow the NumPy docstring format, like:
```
kwargs : dict
    Other Mayavi surface arguments.
```

DingkunLiu: But it is not necessarily to be a dict here. How about changing it to "All additional keyword arguments are passed to the
        `mlab.pipeline.surface` call." and specifying the name `kwargs` to `mlab_kws`?
Or I could just remove the `**` symbols to make it a real dict?

larsoner: Let's do what nibabel does (except we don't need to escape the start characters)

https://github.com/nipy/nibabel/pull/827/files#diff-a1d0ba69368622536897ef9fe4aee05eR766

larsoner: This link is better (sorry on mobile)

https://github.com/nipy/nibabel/blob/ddab5468b12535b044b11837ee2776bb989bc341/nibabel/arraywriters.py#L266

DingkunLiu: There are a lot of detailed arguments for each mayavi function, so I think it is hard to list all the possible arguments. Could we just list some useful examples?

DingkunLiu: And one more thing, I think we should use the same name for the same argument in different functions. For example, the `alpha` in `add_data()` in mayavi is called `opacity`, so I think we should also specify this name in other functions and avoid it to appear in `mlab_kws`.

mwaskom: I don't think we need to list all mayavi keywords, but it would be great if you could enable intersphinx so that the docs produce a link to the relevant mayavi api doc.

Nor do I think that we should have `mlab_kws`; the convention is to collect extra keyword arguments with `**kwargs`, and "mlab" isn't particularly informative about where the kwargs are actually going.

I'd also like to minimize changes unrelated to this specific enhancement. There is some inconsistency between pysurfer parameter names and mayavi ones, but aiming to fully streamline that would be a rabbit hole that is best to avoid.

larsoner: Agreed, I would just do:
```
**kwargs : additional keyword arguments
    These are passed to the underlying :func:`mayavi.mlab.surface` call.
```
To figure out the links you can dump the `intersphinx` inventory with something like:
```
$ python -msphinx.ext.intersphinx http://docs.enthought.com/mayavi/mayavi/objects.inv > mayavi.txt
```
Then browse for relevant entries in `mayavi.txt`. Unfortunately they might not have entries for each function you actually want to use.

DingkunLiu: I have updated the documents, but unfortunately, those for the functions in `mlab.pipeline` are missing.

surfer/viz.py
```diff
@@ -932,6 +932,10 @@ def add_overlay(self, source, min=2, max="robust_max", sign="abs",
             If None, it is assumed to belong to the hemipshere being
             shown. If two hemispheres are being shown, an error will
             be thrown.
+        **kwargs: additional keyword arguments
+            These are passed to the underlying
+                :func:`mayavi.mlab.pipeline.surface` call.
```
larsoner: This extra indentation makes for weird spacing, see CircleCI doc artifact:
```suggestion
            :func:`mayavi.mlab.pipeline.surface` call.
```

https://292-1551431-gh.circle-artifacts.com/0/html/generated/surfer.Brain.html#surfer.Brain.add_foci

And error:

Also many links do not work:
```
/home/circleci/project/surfer/viz.py:docstring of surfer.Brain.add_annotation:37: WARNING: py:func reference target not found: mayavi.mlab.pipeline.surface
/home/circleci/project/surfer/viz.py:docstring of surfer.Brain.add_contour_overlay:42: WARNING: py:func reference target not found: mayavi.mlab.pipeline.contour_surface
/home/circleci/project/surfer/viz.py:docstring of surfer.Brain.add_data:105: WARNING: py:func reference target not found: mayavi.mlab.pipeline.surface
/home/circleci/project/surfer/viz.py:docstring of surfer.Brain.add_foci:41: WARNING: py:func reference target not found: mayavi.mlab.point3d
/home/circleci/project/surfer/viz.py:docstring of surfer.Brain.add_label:43: WARNING: py:func reference target not found: mayavi.mlab.pipeline.surface
/home/circleci/project/surfer/viz.py:docstring of surfer.Brain.add_morphometry:33: WARNING: py:func reference target not found: mayavi.mlab.pipeline.surface
/home/circleci/project/surfer/viz.py:docstring of surfer.Brain.add_overlay:29: WARNING: py:func reference target not found: mayavi.mlab.pipeline.surface
```
These should just be put in double-backticks to render as code. Or do the `intersphinx` dump and find a link that exists in the `objects.inv`.

larsoner: At least one is just a typo, `point3d` should be `points3d` so it directs here

https://docs.enthought.com/mayavi/mayavi/auto/mlab_helper_functions.html#points3d

Fixing these errors is required to get CircleCI happy

DingkunLiu: Sorry for the late reply. I was a little overloaded these days. I just pushed a new commit. And, gladly, the errors seem to be solved.

surfer/viz.py
```diff
@@ -1341,7 +1352,9 @@ def add_label(self, label, color=None, alpha=1, scalar_thresh=None,
             label directory rather than in the label directory itself (e.g.
             for ``$SUBJECTS_DIR/$SUBJECT/label/aparc/lh.cuneus.label``
             ``brain.add_label('cuneus', subdir='aparc')``).
-
+        **kwargs: additional keyword arguments
+            These are passed to the underlying
+            `mayavi.mlab.pipeline.surface` call.
```
larsoner: Can you check all the renderings on CircleCI for errors? For example this one I know will break because you got rid of the section newline -- and indeed the `Notes` section heading is broken:

https://293-1551431-gh.circle-artifacts.com/0/html/generated/surfer.Brain.html#surfer.Brain.add_label

DingkunLiu: Sorry for not very familiar with CircleCI and HTML rendering rules. But actually, the error is in the master branch, see [https://pysurfer.github.io/generated/surfer.Brain.html#surfer.Brain.add_label](https://pysurfer.github.io/generated/surfer.Brain.html#surfer.Brain.add_label). I can try to fix it, but it is not because of my code.

DingkunLiu: See also this [link](https://github.com/sphinx-doc/sphinx/issues/2304).

DingkunLiu: Finally, I figured out how to force line breaking using sphinx...

DingkunLiu: Oh, I thought you meant that the code outside the margin was broken.

surfer/viz.py
```diff
@@ -1339,9 +1350,11 @@ def add_label(self, label, color=None, alpha=1, scalar_thresh=None,
             If a label is specified as name, subdir can be used to indicate
             that the label file is in a sub-directory of the subject's
             label directory rather than in the label directory itself (e.g.
-            for ``$SUBJECTS_DIR/$SUBJECT/label/aparc/lh.cuneus.label``
-            ``brain.add_label('cuneus', subdir='aparc')``).
-
+            for |br| ``$SUBJECTS_DIR/$SUBJECT/label/aparc/lh.cuneus.label``
+            |br| ``brain.add_label('cuneus', subdir='aparc')``).
```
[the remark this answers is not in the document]
DingkunLiu: I just thought the code rendering was an error. As you can see, the code goes out of the margin directly. Do I need to fix this both?

larsoner:
Thanks @DingkunLiu

PR #279, opened 2019-12-10
BUG: Properly clean up on del

larsoner:
This should fix a bug @drammock hit where memory consumption crept up with multiple `fig` use and `mlab.options.offscreen=True`. Basically we should make sure we close the figures we create, since the offscreen ones can stay open and consume memory.

### DIFF
--- a/surfer/viz.py
+++ b/surfer/viz.py
@@ -2298,14 +2298,12 @@ class Brain(object):
         _force_render([])
 
         # should we tear down other variables?
-        if self._v is not None:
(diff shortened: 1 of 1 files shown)


mwaskom:
This looks fine, but has @drammock confirmed that it works? I don’t see a linked issue.

drammock:
Sorry. Yes, confirmed. I probably did so in slack or in a private repo and forgot to report back here.

mwaskom:
Great, thanks @drammock

PR #289, opened 2020-02-26
MRG, FIX: Remove force render

larsoner:
Over at MNE [we have been getting failures](https://github.com/mne-tools/mne-python/pull/7294) that seem to be triggered by the GUI stuff we do when `__del__` is called. This hopefully avoids the problem.

WIP until we can do some runs in MNE to see if it makes the problem go away.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -23,11 +23,11 @@ jobs:
       versionSpec: $(PYTHON_VERSION)
       architecture: $(PYTHON_ARCH)
       addToPath: true
-- task: Cache@2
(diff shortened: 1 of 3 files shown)


larsoner:
Added Azure as a CI while I was at it

larsoner:
Same failures occur in MNE, but I still think this is worth having since the `_force_render` does not seem to be required here

larsoner:
@mwaskom feel free to review/merge if you're happy

mwaskom:
Thanks!

PR #291, opened 2020-03-23
MRG, MAINT: Simpler vector params

larsoner:
We can avoid a lot of gymnastics having to do with maximum magnitudes and even scalars by setting the `scale_mode='vector'` in the `quiver3d` call.


larsoner:
Output unchanged:

https://344-1551431-gh.circle-artifacts.com/0/html/auto_examples/plot_vector_meg_inverse_solution.html#sphx-glr-auto-examples-plot-vector-meg-inverse-solution-py

larsoner:
Pushed a commit that works around Mayavi/VTK auto-scaling stuff

larsoner:
Okay, done for now I think. I tweaked the scaling such that it's more robust to data changes

larsoner:
We are due for a release, so I'll merge this then push out a release unless people object (I think I asked about this already a week or two ago but don't remember...?)

PR #293, opened 2020-05-15
STY: Dont log info messages by default

larsoner:
When processing things in a loop I get spammed by a lot of:
```
colormap sequential: [0.00e+00, 9.68e-01, 1.94e+00] (transparent)
colormap sequential: [0.00e+00, 9.68e-01, 1.94e+00] (transparent)
colormap sequential: [0.00e+00, 9.68e-01, 1.94e+00] (transparent)
colormap sequential: [0.00e+00, 9.68e-01, 1.94e+00] (transparent)
```
I don't think this is very useful for users in general, so by default we probably shouldn't log it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,11 @@ before_install:
     - pip install vtk
     - pip install mayavi
     - mkdir -p $SUBJECTS_DIR
-    - pip install "https://api.github.com/repos/mne-tools/mne-python/zipball/master"
+    - if [ "${PYTHON_VERSION}" == "3.5" ]; then
+        pip install "mne<0.21";
+      else
+        pip install "https://api.github.com/repos/mne-tools/mne-python/zipball/master";
+      fi;
     - python -c "import mne; mne.datasets.fetch_fsaverage(verbose=True)"
 
 install:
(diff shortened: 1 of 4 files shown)


larsoner:
Argh this does not work, I'll probably just set them to `logger.debug` instead of `logger.info`

mwaskom:
+100 on this; I have `verbose=0` set all over the place.

larsoner:
Just changed all `logger.info` to `logger.debug`, then fixed an MNE install error

.travis.yml
```diff
@@ -40,7 +40,11 @@ before_install:
     - pip install vtk
     - pip install mayavi
     - mkdir -p $SUBJECTS_DIR
-    - pip install "https://api.github.com/repos/mne-tools/mne-python/zipball/master"
+    - if [ "${PYTHON_VERSION}" == "3.5" ]; then
```
mwaskom, line 43: We could drop 3.5 support? Most pydata packages have...

larsoner: Let's do it in another PR. It would be good to do that while also adding 3.8 now that VTK9 is out. But that might require quite a few iterations...

PR #294, opened 2020-05-18
ENH: Move up to 3.8

larsoner:
Let's see if it works.

Also bumps minimum Python to 3.6.


larsoner:
Looks like it works! Will merge once CIs come back happy

PR #295, opened 2020-06-08
ENH: Add FXAA option

larsoner:
FXAA can be problematic with OSMESA, give users a way to disable it.


larsoner:
Barring any objections, I'll make CIs happy, merge this and #291, then merge

larsoner:
Sorry, I meant then cut a release

PR #304, opened 2021-07-28
added optional argument 'subjects_dir' to project_volume_data

mmbannert:
Dear PySurfers,

I have improved the "project_volume_data" function in surfer.io. It now accepts an additional optional argument "subjects_dir" of type str (default: None), which overrides the SUBJECTS_DIR environment variable in the call to Freesurfer's mri_vol2surf function by specifying its "--sd" argument.

I find this useful especially when, for instance, the Freesurfer code and the neuroimaging data live on different (remote) servers. 

I have updated the documentation of that function by pasting the explanation of the same argument from PySurfer's "Brain" class. So it should be pretty clear what it does.

I hope you will find this useful!

Cheers,
Michael


agramfort (approved):
LGTM but our CIs are in very bad shape...

@mwaskom @larsoner I think we could merge as is....

larsoner (approved):

PR #320, opened 2023-05-05
Compatibility updates

christianbrodbeck:
This is needed to get PySurfer to import in an environment with Python 3.10 and nibabel 5.1.

### DIFF
--- a/surfer/io.py
+++ b/surfer/io.py
@@ -7,7 +7,10 @@ from subprocess import Popen, PIPE
 import gzip
 import numpy as np
 import nibabel as nib
-from nibabel.spatialimages import ImageFileError
+try:
+    from nibabel.spatialimages import ImageFileError  # removed in nibabel 5.1
+except ImportError:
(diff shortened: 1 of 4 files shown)


larsoner:
MNE-Python no longer uses PySurfer and I don't have the bandwidth to maintain this package. Maybe I could add you as a maintainer @christianbrodbeck ?

christianbrodbeck:
Sure!

larsoner:
Done!

christianbrodbeck:
@Timothy-Anders0n I added that, too, thanks for pointing it out

@larsoner okay to merge this without another review? And how would we go about making a minor release?

larsoner:
Yes you can maintain in whatever way you see fit. Personally I would try to fix CIs then rely on green but that could be a lot of work.

I probably have PyPI credentials to add you so you could upload a sdist and wheel. Let me know your PyPI username if it's different from GitHub

larsoner:
I think I found you on PyPI, invitation sent

